PR #10, opened 2017-07-03
Feature array insert

lahnerml:
Allow inserting of elements at any arbitrary position of an sc_array between 0 and elem_count - 1 and create a test in test_arrays.c


cburstedde:
This function works fine. However,
it has the potentiall to slow the code if used unwisely: each call uses memmove, which will be bad if it is called for a large number of elements. Not merging for now.

PR #12, opened 2019-03-08
Add missing m4 files to be installed w/ libsc

rubendibattista:
Hello, 
in the process of providing packages for p4est in [easybuild](https://easybuild.readthedocs.io/en/latest/) I created two packages: one for `libsc`, and one for `p4est` where the `boostrap.sh` script takes the source dir of `libsc`. 

The problem I encountered is that the install script for `libsc` doesn't install all the macro files in the install prefix `$PREFIX/share/aclocal`. I suppose it is a bug... 

This patch fixes that.


cburstedde:
Thanks for your work on packaging! Much appreciated.

PR #13, opened 2019-03-11
Ref count bypass

n8xm:
Added ability to skip reference counting in `sc_malloc`, `sc_calloc` and `sc_free` if `DSC_NOCOUNT_MALLOC` is defined. This is useful when multithreading without using `--enable-pthread`, because without using pthreads the aforementioned functions are not thread safe.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -80,7 +80,7 @@ echo "o---------------------------------------"
 echo "| Checking functions"
 echo "o---------------------------------------"
 
-AC_CHECK_FUNCS([backtrace backtrace_symbols strtol strtoll])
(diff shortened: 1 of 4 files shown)


n8xm:
Closing because there seem to be changes other than my own (from some other branch) in this pull request.

n8xm:
Besides fixing the aforementioned problem with this pull request, I need to test on another machine and and commit a `doc/author_lastname.txt` in order to "make my code BSD". I expect to follow up with another pull request.

cburstedde:
> Besides fixing the aforementioned problem with this pull request, I need to test on another machine and and commit a `doc/author_lastname.txt` in order to "make my code BSD". I expect to follow up with another pull request.

Thanks.  Probably best to leave sc_psort to a different issue.

PR #14, opened 2019-03-12
Add define that disables reference counters

n8xm:
Added ability to skip reference counting in `sc_malloc`, `sc_calloc` and `sc_free` if `SC_NOCOUNT_MALLOC` is defined. This is useful when multithreading without using `--enable-pthread`, because without using pthreads the aforementioned functions are not thread safe.

### DIFF
--- a/doc/author_marshak.txt
+++ b/doc/author_marshak.txt
@@ -1,0 +1,1 @@
+I hereby place my contributions to libsc under the FreeBSD license. - Nathan Marshak
--- a/src/sc.c
+++ b/src/sc.c
@@ -534,12 +534,16 @@ sc_malloc (int package, size_t size)
 #ifdef SC_ENABLE_PTHREAD
   sc_package_lock (package);
 #endif
+
+#ifndef SC_NOCOUNT_MALLOC
   if (size > 0) {
     ++*malloc_count;
   }
   else {
     *malloc_count += ((ret == NULL) ? 0 : 1);
   }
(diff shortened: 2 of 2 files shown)


cburstedde:
> Added ability to skip reference counting in `sc_malloc`, `sc_calloc` and `sc_free` if `SC_NOCOUNT_MALLOC` is defined. This is useful when multithreading without using `--enable-pthread`, because without using pthreads the aforementioned functions are not thread safe.
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/cburstedde/libsc/pull/14
> 
> -- Commit Summary --
> 
>   * Bypass memory counters if SC_NOCOUNT_MALLOC is passed
>   * Licesned my contribution under BSD
> 
> -- File Changes --
> 
>     A doc/author_marshak.txt (1)
>     M src/sc.c (12)
> 
> -- Patch Links --
> 
> https://github.com/cburstedde/libsc/pull/14.patch
> https://github.com/cburstedde/libsc/pull/14.diff

Thanks, I've added two more #defines SC_NOCOUNT_REFCOUNT and
SC_NOCOUNT_LOGINDENT.  If you could use all three, we should be safe for a
while.

n8xm:
I think we are done here - for future reference, see the [fix-bypass-memcounters](https://github.com/cburstedde/libsc/tree/fix-bypass-memcounters) branch. Closing.

PR #16, opened 2019-07-10
test/test_dmatrix - Increase tolerance and make test more robust

tamiko:
It turns out that on non-amd64 architectures the test test_dmatrix fails
due to differences in the compared matrices larger than floating point
roundoff errors [1]. Changing the limit from DBL_MIN to 1.0e4 *
DBL_EPSILON [2] (assuming that differences are indeed of the order of
eps) fixes the problem .
    
[1] https://launchpad.net/~ginggs/+archive/ubuntu/testing/+sourcepub/10330424/+listing-archive-extra
[2] https://en.cppreference.com/w/c/types/limits

### DIFF
--- a/test/test_dmatrix.c
+++ b/test/test_dmatrix.c
@@ -73,7 +73,7 @@ test_dmatrix_check_error_identical (const sc_dmatrix_t * mat_chk,
   SC_ASSERT (totalsize == mat_ref->m * mat_ref->n);
 
   for (i = 0; i < totalsize; ++i) {
-    if (DBL_MIN < fabs (mat_chk_data[i] - mat_ref_data[i])) {
+    if (1.e4 * DBL_EPSILON < fabs (mat_chk_data[i] - mat_ref_data[i])) {
       error_count++;
     }
   }


cburstedde:
Thanks! Can you please commit a file doc/author_maier.txt to put your changes to libsc either into the public domain or under the FREEBSD license?

cburstedde:
Hi, bumping this for the requested author text file. Thanks!

holke:
Hi.
I can confirm that the dmatrix test fails on an Intel i7-6600U when configured with the C flag 
```
-march=native
```
Strangely though, it only occurs when running 
```
make -j check
``` 
and not with 
```
make check
```
(any idea why?)

This fix resolves the error.

cburstedde:
The dependency on -j is weird, but addressed by the fix.

I'll merge it but I cannot do that until I the author_tamiko.txt file is added to put the code under the FreeBSD license.

cburstedde:
> I'll merge it but I cannot do that until I the author_tamiko.txt file is added to put the code under the FreeBSD license.

Hi can I ping you again about this little file? Thanks!

cburstedde:
Thanks for your proposal! I'm closing this PR though since I don't have the author file.

PR #17, opened 2020-02-22
Disable SIGUSR2 signal on MinGW platforms

sloede:
I was trying to compile p4est on MinGW and this was basically the only issue that prevented me from doing so. Therefore I opened this PR to see if you would be willing to include this in libsc.

### DIFF
--- a/src/sc.c
+++ b/src/sc.c
@@ -27,6 +27,11 @@
 #include <signal.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


cburstedde:
Thanks, I've just removed SIGUSR2 from the code entirely. See latest develop branches.
I appreciate your contribution! Please don't worry about the closed pull request.

PR #19, opened 2020-04-05
sc_psort issue

haghakhani:
there are two commits in the new branch:

1) making minor changes and activate test/test_sort.c -->  the changes activate the test and with some minor polish. The test is fine and sc_psort works as expected. 

2) adding time and date to the logger --> this commit adds date and time to the logger. I found it very useful when I run a job in a cluster. This gives a relatively good execution timing and in cases that the job fails may help in finding the cause by tracing back the time. The output would be like the following:
[libsc 7] **[2020-4-4 23:46:43]**     47.9      48    48.1    48.3    48.5    49.3    49.8

### DIFF
--- a/src/sc.c
+++ b/src/sc.c
@@ -45,6 +45,10 @@ typedef void        (*sc_sig_t) (int);
 #include <pthread.h>
 #endif
 
+#ifdef SC_HAVE_TIME_H
+#include <time.h>
+#endif
+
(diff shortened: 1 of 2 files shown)


cburstedde:
Thanks again, this branch is now merged into develop and prev3-develop. Will eventually become part of master!

PR #25, opened 2020-11-18
Feature version functions

ranocha:
This is a follow-up of #24. I implemented a function `sc_version_point` as discussed earlier today, ran `scindent` (discarding changes to `__attribute__`), and verified that tests pass locally (including the new one).

### DIFF
--- a/src/sc.c
+++ b/src/sc.c
@@ -39,9 +39,6 @@ typedef void        (*sc_sig_t) (int);
 #endif
 #endif
 
-#include <errno.h>
-#include <stdlib.h>
-
 #ifdef SC_ENABLE_PTHREAD
 #include <pthread.h>
 #endif
@@ -1469,14 +1466,15 @@ sc_version_minor (void)
 }
 
 /* Define helper macros temporally to convert the point version */
-#define _TOSTRING(x) #x
-#define TOSTRING(x) _TOSTRING(x)
+#define _SC_TOSTRING(x) #x
+#define SC_TOSTRING(x) _SC_TOSTRING(x)
 
 int
 sc_version_point (void)
 {
-  /* SC_VERSION_POINT may contain a dot and/or dash, followed by additional information */
-  return strtol (TOSTRING (SC_VERSION_POINT), NULL, 10);
+  /* SC_VERSION_POINT may contain a dot and/or dash,
+     followed by additional information */
+  return atoi (SC_TOSTRING (SC_VERSION_POINT));
 }
 
 #undef TOSTRING
--- a/src/sc.c
+++ b/src/sc.c
@@ -40,6 +40,7 @@ typedef void        (*sc_sig_t) (int);
 #endif
 
 #include <errno.h>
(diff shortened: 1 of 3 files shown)


cburstedde:
Nice work!

ranocha:
Thanks! I didn't use `atoi` since I wasn't sure whether "undefined behavior" (possibly including [nasal demons](http://catb.org/jargon/html/N/nasal-demons.html)) is acceptable in `libsc`. It looks like it is and I will try to keep that in mind. Thanks for adjusting this PR!

src/sc.c
```diff
@@ -1467,4 +1465,19 @@ sc_version_minor (void)
...
+int
+sc_version_point (void)
+{
+  /* SC_VERSION_POINT may contain a dot and/or dash,
+     followed by additional information */
+  return atoi (SC_TOSTRING (SC_VERSION_POINT));
+}
+
+#undef TOSTRING
+#undef _TOSTRING
```
ranocha, line 1481: I didn't want to clutter the namespace with macros that are not used elsewhere. Hence, I `#undef`ed the macros above after their use in `sc_version_point`. Since you changed the name to include `SC_`, the `#undef` statements should probably be updated, too. What do you suggest, @cburstedde?

cburstedde: Uh yes you're right on both counts.

Undefined behavior is generally not acceptable. We get around it this time since we have a test program. Let's see what happens if we don't have a point, say checking out v2.2. Will the test go through?

I didn't see the undef of TOSTRING. I'd tend to remove it. Still we may move the macro definitions to sc.h.

ranocha: Okay. Shall I do some of these steps or is it just easier for you to do that?

cburstedde:
> +#undef TOSTRING
> +#undef _TOSTRING
> 
> Okay. Shall I do some of these steps or is it just easier for you to do that?

Please feel free to go ahead, and to update the test to sc3-type licence.

ranocha:
> Please feel free to go ahead, and to update the test to sc3-type licence.
> 

Okay, I'll do that tomorrow.

cburstedde:
By reading the man pages of atoi and strtol a couple times, I did see that atoi will not detect errors. It does not say anywhere that *behaviour* is undefined. The return value for no digits is defined as 0 when endptr is supplied to strtol, but I do not see a defined return *value* when no digits are found. In any case I'd think we are safe from demons... opinions?

ranocha:
I found the following description of `atoi` on http://www.cplusplus.com/reference/cstdlib/atoi/

> If the converted value would be out of the range of representable values by an int, it causes undefined behavior. See strtol for a more robust cross-platform alternative when this is a possibility.

But that's just a minor problem and shouldn't get us into trouble here, since `atoi` is used in a very controlled environment, acting on the point version and not any user input. Hence, it should be fine to use it here.

ranocha:
> Please feel free to go ahead, and to update the test to sc3-type licence.

Done in #26.

PR #26, opened 2020-11-21
Feature version functions

ranocha:
As discussed in https://github.com/cburstedde/libsc/pull/25, I've
- moved the definition of `SC_TOSTRING` to `sc.h`
- updated the license header of `test/test_version.c`

### DIFF
--- a/src/sc.c
+++ b/src/sc.c
@@ -1465,10 +1465,6 @@ sc_version_minor (void)
   return SC_VERSION_MINOR;
 }
 
-/* Define helper macros temporally to convert the point version */
-#define _SC_TOSTRING(x) #x
-#define SC_TOSTRING(x) _SC_TOSTRING(x)
-
 int
 sc_version_point (void)
 {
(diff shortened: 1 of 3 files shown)


cburstedde:
So I experimented with version tags. The following pops up in src/sc_config.h when the version is set with git tag -s (git describe works ok):

 * v7 major = minor = point = 7
 * v8.9 major = 8 minor = 9 point = 8.9

The workaround for the first is not to use v7 but the v7.0 convention instead.
Considering the second, the point version is unreliable as identified by (my) autoconf.
Both is sad. To keep things simple, let's just not use point as a number, which anyway was the original suggestion by @ranocha.
I'll go push these couple changes to this PR while I'm at it.

Since we won't need atoi anymore for this PR, nasal demons will be discussed elsewhere. I'm officially afraid of them now.

ranocha:
That sounds good to me. I will remove the function `p4est_version_point` from the related PR https://github.com/cburstedde/p4est/pull/87.

PR #28, opened 2021-01-26
CMake build system (speed, many OS and compilers)

scivision:
This allows using CMake, which configures and builds libsc in < 10 seconds instead of ~ 10 minutes on Windows Subsystem for Linux with autotools.

Tested on 

* MacOS 10.15: gcc-10, intel oneAPI 2021.1, Clang
* Linux: gcc-9, intel oneAPI 2021.1
* Windows: gcc-10, intel oneAPI 2021.1 in Clang mode


cburstedde:
Thanks, I'd be curious to try this out. May I ask you to provide a file doc/author_lastname.txt placing your code under BSD?

I'd prefer to take out the M_PI define and to make it a sleep(1) instead of 3. Let me also verify the basename usage.

cburstedde:
Out of curiosity, have you ever compiled it and run checks on 32 bit?

cburstedde:
Ah and we'd likely want to keep the blas/lapack/fortran configure completely out of prev3-develop (not needed in the future).

scivision:
The M_PI, basename and sleep workarounds are to make it work with MSVC-based compilers without having to create functions to abstract those away (using the same techniques).

cburstedde:
> The M_PI, basename and sleep workarounds are to make it work with MSVC-based compilers without having to create functions to abstract those away (using the same techniques).

That's cool. For p4est we'd need M_PI_4, too.

scivision:
yes the blas was a leftover copying mistake, I removed it

scivision:
I didn't need M_PI after all, I forgot to put in a `#define` right before math.h in sc.h.

scivision:
I have built the whole forestclaw + p4est + sc on 32-bit ARM if that's what you mean, using CMake

cburstedde:
> I have built the whole forestclaw + p4est + sc on 32-bit ARM if that's what you mean, using CMake

Coll, I have this report cburstedde/p4est#63 about one test failing on 32bit.
Is that reproducible?

cburstedde:
I've renamed the branch to feature-cmake on my end and added a few edits.

My CMake is at version 3.13.4. is it possible to relax the minimum version to work a couple years backwards?

scivision:
Yes the techniques I'm using coincidentally require CMake >= 3.13 so I can change a few commands and it will work with 3.13.

scivision:
I have switched p4est and libsc to CMake >= 3.13

cburstedde:
> I have switched p4est and libsc to CMake >= 3.13

Thanks!  One thing I noticed, on my linux box it did not detect qsort_r, which
is found by autoconf.

scivision:
[arm.log](https://github.com/cburstedde/libsc/files/5889833/arm.log)

that's from a Raspberry Pi 4 with 2GB RAM,  tests/examples run using `mpiexec -n 4`

scivision:
qsort_r is fixed

cburstedde:
Thanks for checking on qsort_r. This one has two different call conventions, it seems BSD differst from GNU, I have another pending conversation about how to go about this on the autotools side.

Did you get a chance to look at the commits I had pushed in between? The status I had put together is at ab304b8004f93b1f1169b52af04b8503bb1b8fec, and I'm calling the branch feature-cmake.

scivision:
I incorporated your `feature-cmake` changes here

scivision:
The changes I made to the .c source code were just to work on Windows. I had to make them part of this PR as I couldn't test CMake on Windows without them.

I think this will also work on additional compilers, but for time constraints I only tested on the most commonly used today.

cburstedde:
> The changes I made to the .c source code were just to work on Windows. I had to make them part of this PR as I couldn't test CMake on Windows without them.

Yes, I understand.  I just updated them to remove unititialized pointer references.

> I think this will also work on additional compilers, but for time constraints I only tested on the most commonly used today.

Cool, that's definitely more compilers that we have tested so far.

cburstedde:
> I incorporated your `feature-cmake` changes here

Can I ask you to have them begin on top of ab304b8004f93b1f1169b52af04b8503bb1b8fec?

cburstedde:
I've merged our two feature-cmake branches; the new tip is b37f8431f4e269cb89981b09dd28db3a4b982316.

cburstedde:
I've merged feature-cmake into prev3-develop and forwarded feature-cmake to the same tip.

I've renamed the config templates to src/sc_config_{autotools,cmake}.h.in, seems to be working.
I'll continue to build the tarball with autotools' make dist. I've testet the cmake files to be included.

If there's more code to be pushed, I'd be happy to look at future PRs on any feature-cmake branch.
Thanks again!

scivision:
Thanks looks great!

PR #29, opened 2021-02-10
cmake: CI and install/package

scivision:
# Add CI and install package

* build example/ separately to demonstrate finding / linking libsc

## Packaging libsc

This command uses CMake to make a .zip ready to distribute to machines of similar OS, libc and CPU arch:

Assuming you have as usual:

```sh
cmake -B build
cmake --build build --parallel
```

Create the package by:

```sh
cpack --config build/CPackConfig.cmake
```

For example, on Windows this creates SC-0.0.0-win64.zip where 0.0.0 is the libsc version number set in CMakeLists.txt `project(VERSION)`

The options for the package are set at the bottom of cmake/install.cmake.

For example, this package can be distributed automatically via GitHub Releases with a couple more parameters added to .github/workflows/ci.yml

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,20 +16,12 @@ jobs:
       uses: actions/checkout@v2
     - name: Run bootstrap script
       run: ./bootstrap
-    - name: Make check with debug, without BLAS and shared
+    - name: Make check with MPI and debug, without shared
       shell: bash
       run: |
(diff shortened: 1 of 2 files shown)


scivision:
It looks like you might need to enable GitHub Actions. This is a one-time configuration at: https://github.com/cburstedde/libsc/actions

Then you'll get a dashboard linking to results like: https://github.com/scivision/libsc/actions/runs/555720038

PR #30, opened 2021-02-12
Add CMake CI, improve packaging

scivision:
These were accidentally dropped from the prior merge. All is well, they simply go on top of the prior commits.

A CI workflow for CMake is relevant as other CMake packages build on this, it's better to have each package with CMake CI to aid troubleshooting.


cburstedde:
Works nicely. Thanks again!

PR #32, opened 2021-02-14
CMake self-test, presets, improve CI and config checks

scivision:
CMake improvements realized during integration with p4est.

* Ensure that self-tests do not encounter intermittent errors due to inter-test MPI contention by CMake RESOURCE_LOCK
* add configure-time checks that sc_config.h was generated correctly, and also check this in FindSC.cmake
* add CMakePresets.json, an optional file that enables users to declare arbitrarily complex system configurations
* improve CI coverage using presets
* add Windows MS-MPI to CI
* make CMake compiler options more like autotools and add more compilers


cburstedde:
Thanks again! When I run cpack on linux, it builds sc-0.0.0-Source.tar.gz. Would it be possible to align the archive name to the previous convention `libsc-2.8.2.tar.gz`, `libsc-2.8.2.1-277b.tar.gz` derived from `git describe`?

scivision:
yes do you want the plain `git describe` or any additional `--abbrev`?

cburstedde:
> yes do you want the plain `git describe` or any additional `--abbrev`?

I'd say sticking to the default will be fine.  I've been using --dirty --tags
in the CI and that seems to be ok.

cburstedde:
> yes do you want the plain `git describe` or any additional `--abbrev`?

So we have been using --abbrev=4.  Let's keep that...

PR #33, opened 2021-02-16
cmake: add pkg-config generate and determine version from git describe

scivision:


scivision:
that error code 135 on Windows CI is just something that happens intermittently on GitHub Actions with Windows in general. I.e. just rerunning the workflow clears it

cburstedde:
Thanks! I'm just testing this and I'm getting on `cmake path/to/source/p4est/sc`:
(This seems to be independent of updating `git describe --abbrev=4 --match 'v*'`.)

```
-- SC 2.8.2
-- NAME_MISMATCHED (missing: REQUIRED_VARS VERSION_VAR) 
CMake Error at /usr/share/cmake-3.13/Modules/FindPackageHandleStandardArgs.cmake:195 (message):
  Unknown keywords given to FIND_PACKAGE_HANDLE_STANDARD_ARGS():
  "REASON_FAILURE_MESSAGE"
Call Stack (most recent call first):
  cmake/Modules/FindMPI.cmake:1741 (find_package_handle_standard_args)
  cmake/config.cmake:8 (find_package)
  CMakeLists.txt:21 (include)
```

scivision:
I'm going to remove that backported FindMPI.cmake, as it causes issues with older CMake

cburstedde:
> I'm going to remove that backported FindMPI.cmake, as it causes issues with older CMake

Sounds great.  My preference is simplicity always.

scivision:
OK I think this is ready to go, thanks

cburstedde:
Cool, thanks! As a side note, we never had much luck with the github release mechanism, so I'm not endorsing github releases. Especially the p4est/sc submodule never was included properly.

cburstedde:
Testing this, I find a couple differences with autotools regarding the .pc:
 * The file name would ideally look like libsc-2.8.2.25-afe3-dirty.pc
 * Can .pc file name and "Version:" be the same as by git describe --dirty --abbrev=4 --match="v*"?
 * Can we remove the "Requires: sc" line?
 * I'd prefer not to expose iniparser at all, not even privately, just to add the objects to the library.
 * (Note to self: Make sure -lsc shows up in the autotools .pc, and quote the variables.)

scivision:
Do you also want libb64 to also be just an object like iniparser?

cburstedde:
> Do you also want libb64 to also be just an object like iniparser?

That would be great, then .pc: Libs.private becomes the same as Libs and might no longer be needed.

The default --prefix for autoconf is $PWD/local, would this be an easy thing to do with cmake?

I'm about to integrate up to the current tip, please don't force push at this point. Thanks again!

cburstedde:
Is there anything I should make install in the autotools build (such as cmake related files)?

scivision:
In the next PR I put all non-CMakeLIsts.txt Cmake files under cmake/

PR #38, opened 2021-03-16
Make SC3_REFCOUNT_MAGIC long as it is used in code

mkirilin:
Typecast  of SC3_REFCOUNT_MAGIC

GNU 10.2.0 on Windows treated SC3_REFCOUNT_MAGIC as long long type, while the type of sc3_refcount_t::magic is long.
The result of comparison r->magic == SC3_REFCOUNT_MAGIC might have varied depending on compiler / operational system / device.


cburstedde:
Thanks, good catch!

PR #41, opened 2021-04-02
Feature cmake

mkirilin:
MPI and OpenMP are disabled by default. Use -Dmpi=yes and -Dopenmp=yes to enable them.
Also, ci runs (for examples) were changed according to new defaults.


cburstedde:
We've merged this outside of github, thanks for the contribution.

cburstedde:
Ok this one is not redundant at all since it applies to libsc.

cburstedde:
Thanks! Will update p4est/sc submodule on my end.

PR #42, opened 2021-04-29
CMake: improve install

scivision:
Add findSC to install so that consuming programs will be able to find libsc.

cleanup unused variables and improve config logic.


cburstedde:
Thanks! Merged.

For remaining open issues concerning cmake please cf. cburstedde/p4est#103, I'm trying to collect all I can think of for whoever feels like working on it.

PR #43, opened 2021-05-12
sc3: array: Add a view option into sc3_array

mkirilin:
# sc3 array view

I added an opportunity to make subarray/view to sc3_array_t. It's necessary to call sc3_array_split() on p4est3 quadrants array and might be useful for future tasks.

### DIFF
--- a/src/sc3_array.c
+++ b/src/sc3_array.c
@@ -38,6 +38,7 @@ struct sc3_array
 
   /* parameters fixed after setup call */
   int                 initzero, resizable, tighten;
+  int                 view;
   int                 ecount, ealloc;
   size_t              esize;
 
@@ -220,7 +221,7 @@ sc3_array_unref (sc3_array_t ** ap)
     *ap = NULL;
 
     aator = a->aator;
-    if (a->setup) {
+    if (a->setup && !a->view) {
       /* deallocate element storage */
       SC3E (sc3_allocator_free (aator, a->mem));
     }
@@ -363,6 +364,42 @@ sc3_array_index_noerr (const sc3_array_t * a, int i)
   }
 #endif
   return a->mem + i * a->esize;
+}
+
+sc3_error_t        *
+sc3_array_subarray (sc3_array_t * a, int idx, sc3_array_t * sa)
+{
+  SC3A_IS (sc3_array_is_setup, a);
+  SC3A_IS (sc3_array_is_new, sa);
+  SC3A_CHECK (sa->ecount + idx <= a->ecount);
+  SC3A_CHECK (sa->esize == a->esize);
+  SC3A_IS (sc3_array_is_unresizable, sa);
+
+  sa->view = 1;
+  sa->setup = 1;
+  sa->initzero = a->initzero;
+  sa->tighten = 0;
+
+  sa->ealloc = a->ealloc - idx;
+  sa->mem = a->mem + idx * sa->esize;
+
+  return NULL;
+}
+
+sc3_error_t        *
+sc3_array_view (char * a, int idx, int n, sc3_array_t * sa)
+{
+  SC3A_IS (sc3_array_is_new, sa);
+  SC3A_IS (sc3_array_is_unresizable, sa);
+
+  sa->view = 1;
+  sa->setup = 1;
+  sa->tighten = 0;
+
+  sa->ealloc = n;
+  sa->mem = a + idx * sa->esize;
+
+  return NULL;
 }
 
 sc3_error_t        *
--- a/src/sc3_array.c
+++ b/src/sc3_array.c
@@ -367,37 +367,46 @@ sc3_array_index_noerr (const sc3_array_t * a, int i)
 }
 
 sc3_error_t        *
-sc3_array_subarray (sc3_array_t * a, int idx, sc3_array_t * sa)
+sc3_array_new_view (sc3_array_t * a, int offset, int length,
+                    sc3_array_t ** view)
 {
   SC3A_IS (sc3_array_is_setup, a);
-  SC3A_IS (sc3_array_is_new, sa);
-  SC3A_CHECK (sa->ecount + idx <= a->ecount);
-  SC3A_CHECK (sa->esize == a->esize);
-  SC3A_IS (sc3_array_is_unresizable, sa);
+  SC3A_IS (sc3_array_is_new, *view);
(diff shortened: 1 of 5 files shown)


cburstedde:
Thanks, shall we use a naming convention similar to the existing sc_containers:
sc3_array_new_view (views an existing array)
sc3_array_new_data (the *dangerous* memory-view function, which should also take an element size)
Can both functions take an additional length parameter? They'd also take an sc3_array_t ** output (first) argument.
Any other issues to discuss?

mkirilin:
Thank you for the remarks. I've updated the interface.
I also kept an _offset_ variable for a new_data function. I suppose, the only thing left to discuss about sc3_array_t is the way of returning value from a pop function.

cburstedde:
Thanks, I've made some adjustments, basically calling `sc3_array_new` inside the new functions.

I'm suspecting that `sc3_array_is_valid` will not go through with all views. Would you like to write some tests for the new functions?

mkirilin:
I've added some tests and fix the bug they've uncovered. I also noted that it is important to destroy a view before an array; added this to the documentation.

src/sc3_array.c
```diff
@@ -406,13 +407,16 @@ sc3_array_new_view (sc3_allocator_t * alloc, sc3_array_t ** view,
...
   (*view)->ecount = length;
   (*view)->mem = a->mem + (*view)->esize * offset;
 
   (*view)->viewed = a;
   SC3E (sc3_array_ref (a));
 
+  lg = sc3_log2_ceil (length, SC3_INT_BITS - 1);
+  SC3A_CHECK (0 <= lg && lg < SC3_INT_BITS - 1);
+  SC3A_CHECK (length <= (1 << lg));
+  (*view)->ealloc = 1 << lg;
```
cburstedde, line 419: I think this power of 2 requirement is articial/historic and not useful for views (especially those on data). Let's just assign ealloc = length for both view constructions and update the test in sc3_array_is_valid for views accordingly.

P.S. Thanks for the tests!

mkirilin:
I got rid of power of 2 necessity in views.
Added a new option to renew view / data without memory reallocation. It's possible to convert view to data and vice versa via the renew function now.
Also, edit tests for a bit to test renew concept.

cburstedde:
> I got rid of power of 2 necessity in views.

Thanks, looks good!

> Added a new option to renew view / data without memory reallocation. It's possible to convert view to data and vice versa via the renew function now.
> Also, edit tests for a bit to test renew concept.  

I'm wondering how to make this a bit stricter and still well usable.

What do you think about restricting renew_view to views onto arrays and
renew_data to views on data in order to forbid future mixups.

We might also begin array views in resizable status and allow for calling
sc3_array_resize on them, unless the view is frozen explicitly.  For
consistency, we should default normal arrays to resizable as well.

What do you think?

cburstedde:
I had tweaked the documentation content a bit for new_view, _data to reformulate the destroy requirement.
Could you adjust along these lines?

mkirilin:
> What do you think about restricting renew_view to views onto arrays and
> renew_data to views on data in order to forbid future mixups.

I think, that the absence of the restriction makes view more convenient instrument for a user. I suppose, it would be a good compromise to add some SC3_IS-compatible functions that would check if a view track an array or data at the moment.

> We might also begin array views in resizable status and allow for calling
> sc3_array_resize on them, unless the view is frozen explicitly.  For
> consistency, we should default normal arrays to resizable as well.

I didn't get. Do you mean, we need to redefine an original array as resizable, even if  the array was defined unresizable  by a user? What if we leave the original array the same as it previously defined and allow the view to be resizable, but restrict it  by array's boundary?

> I had tweaked the documentation content a bit for new_view, _data to reformulate the destroy requirement.
> Could you adjust along these lines?
Yes, sure.

cburstedde:
> > What do you think about restricting renew_view to views onto arrays and
> > renew_data to views on data in order to forbid future mixups.
> 
> I think, that the absence of the restriction makes view more convenient instrument for a user. I suppose, it would be a good compromise to add some SC3_IS-compatible functions that would check if a view track an array or data at the moment.

Let's do this and add sc3_array_is_data, making is_view false for views on
data.  I'd like to have the renew functions maximally fast outside of debug
mode, so changing the type of view may not be what we want. 

> > We might also begin array views in resizable status and allow for calling
> > sc3_array_resize on them, unless the view is frozen explicitly.  For
> > consistency, we should default normal arrays to resizable as well.
> 
> I didn't get. Do you mean, we need to redefine an original array as resizable, even if  the array was defined unresizable  by a user? What if we leave the original array the same as it previously defined and allow the view to be resizable, but restrict it  by array's boundary?

Yes, that's exactly what I meant.  The original array must stay the same.  The
resize function may be inconvenient for a view though, since it can specify a
length but not an offset.  So let's better not touch it.

It is always possible to use renew_view with the same array as before, the code
should allow this to be handled quickly (skipping the unref and ref).

> > I had tweaked the documentation content a bit for new_view, _data to reformulate the destroy requirement.
> > Could you adjust along these lines?
> 
> Yes, sure.

Thanks!

mkirilin:
>  > What do you think about restricting renew_view to views onto arrays and
>  > renew_data to views on data in order to forbid future mixups.
> 
>  I think, that the absence of the restriction makes view more convenient instrument for a user. I suppose, it would be a good compromise to add some SC3_IS-compatible functions that would check if a view track an array or data at the moment.
> Let's do this and add sc3_array_is_data, making is_view false for views on
> data.  I'd like to have the renew functions maximally fast outside of debug
> mode, so changing the type of view may not be what we want.

Yes, I agree with you about necessity of fast execution of renew_view functions. It does make sense to restrict  renew functions to view / data.

>  > We might also begin array views in resizable status and allow for calling
>  > sc3_array_resize on them, unless the view is frozen explicitly.  For
>  > consistency, we should default normal arrays to resizable as well.
> 
>  I didn't get. Do you mean, we need to redefine an original array as resizable, even if  the array was defined unresizable  by a user? What if we leave the original array the same as it previously defined and allow the view to be resizable, but restrict it  by array's boundary?
> Yes, that's exactly what I meant.  The original array must stay the same.  The
> resize function may be inconvenient for a view though, since it can specify a
> length but not an offset.  So let's better not touch it.

We can calculate a subtraction between pointers points to begin array's and view's memory to dedicate an offset.

PR #44, opened 2021-06-01
Feature sc3 array split

mkirilin:
Added sc3_array_split just as in p4est2 but with p4est3 interface. Also, I added sc3_array_is_sorted function.


mkirilin:
I removed types changes from the history. No one will know about it and won't be mislead by that.

mkirilin:
I rolled back declaration of compar function prototype in _is_sorted.

PR #47, opened 2021-06-30
Feature timeofday

gghosh8495:
# Title

Following up on issue # .

Proposed changes: Added struct timezone

### DIFF
--- a/doc/author_ghosh.txt
+++ b/doc/author_ghosh.txt
@@ -1,0 +1,3 @@
+I place my contributions to libsc under the FreeBSD license. 
+Gautam Ghosh (ghosh@ins.uni-bonn.de)
+


gghosh8495:
@cburstedde  I am not sure if the diff will help you understand but as you wanted I have put the un-indented version and the indented version. 
If the diff is not clear then I would be happy to explain or put new commits.

Note that the develop branch does not have timeofday,c. I copy-pasted the new code entirely.

gghosh8495:
@cburstedde  The change is minimal, we just need to define struct timezone and pass it to the function. We use that if the pointer is valid. That is what I understand from https://stackoverflow.com/a/26085827

cburstedde:
Thanks, sorry I'm late getting to this: In the prev3-develop branch there is already a gettimeofday function. Could we just use that one or what would you propose choosing?

PR #48, opened 2021-07-02
timeofday

gghosh8495:
# Title Fix for getimeofday pointer visibility 

Following up on issue # .

Proposed changes: Define struct timeval and pass it to the function

### DIFF
--- a/src/sc_builtin/gettimeofday.c
+++ b/src/sc_builtin/gettimeofday.c
@@ -2,45 +2,48 @@
 
 #define WIN32_LEAN_AND_MEAN
 #include <Windows.h>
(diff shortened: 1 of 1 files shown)


gghosh8495:
@cburstedde Sorry for the delay, I ran into some unexpected issues with the diff and text editor. You can see the commit id f781af8 for semantic changes. The top commit is indented and introduces line shifts in diff.

cburstedde:
Thanks, looks good!

cburstedde:
One more thing: Could you please make sure gettimeofday.c gets added to the autotools make dist result?

PR #50, opened 2021-09-09
fabs check for CMake

gghosh8495:
# Title Checking for fabs function in CMake 

Following up on issue # .

Proposed changes:
I know the SC_HAVE_FABS is not really required, I am adding that just for the sake of consistency with Autoconf.


cburstedde:
Looks good!

PR #54, opened 2022-02-13
CMake: Bugfix: Do not compile and link libb64/b64(dec|enc).c

tamiko:
Both source files only contain a single main() function that adds no
functionality to a shared or static library. Worse, newer linkers
start to reject linking object files with multiply defined global
symbols (in this case "main").

Simply remove both source files from the internal object library.

Ps.: I am not sure against what feature branch / developing branch I am
supposed to make a pull request. I simply chose "feature-cmake".


cburstedde:
Yes, good call, thanks.

PR #56, opened 2022-02-16
Macro interference for builds

gghosh8495:
# Title

Following up on issue # .

Proposed changes:
For old C functions like qsort_r, snprintf, vsnprintf you need to #define _GNU_SOURCE because functions like these are not a part of ISO C99 standard.


gghosh8495:
The windows build is a weird issue, that needs a bit more time to fix. Linux and macOS are good to go.

cburstedde:
Thanks: could you please squash the history such that there is one remaining change per commit? Should be just a few. Can you make the _MSC_VER check look the same in sc.h and sc3_base.h?

PR #58, opened 2022-02-18
CMake: add v4l2 like autotools

scivision:
The new v4l2 wasn't in CMake, so this adds it and the v4l2 example.
Relevant symbols are added to sc_config.h.

cmake/options.cmake adds parameters relevant for shared build of libsc--enabling Rpath for Unix-like and on Windows exporting symbols for .lib import library that accompanies .dll.


cburstedde:
Thanks for aligning cmake with the autotools.

PR #59, opened 2022-02-18
CMake: improve shared test and build

scivision:
Add further options, particularly for Windows, enabling proper build and packaging of shared library libsc.
add CI tests for shared/static on all platforms.


cburstedde:
Hi did I get the merge right?

cburstedde:
Thanks!

PR #60, opened 2022-02-24
CI:CMake: further compiler tests and improve efficiency

scivision:
Speed up CI by removing unneeded Ninja install (Make or Ninja works fine).
Speed up Windows CI by using default image GCC
Add Clang test to MacOS

Enable SC debug flags also for CMake RelWithDebInfo builds, which uses flags like `-O2 -g`


cburstedde:
Thanks! As always I'll welcome whatever trickles down to p4est and forestclaw in a similar way.

PR #61, opened 2022-02-24
cmake:findSC: use canonical CMake approach to find/read sc_config.h

scivision:
this avoids polluting CMakeError.log with nuisance messages when
SC is not completely found.

(cherry picked from p4est PR)


cburstedde:
Thanks again for working on our CMake!

PR #68, opened 2022-03-09
cross platform compatibility for M_PI

gghosh8495:
# Title

Following up on issue # 66.

Proposed changes:
Explicitly #define M_PI constants for compatibility with older C compilers.

### DIFF
--- a/src/sc.h
+++ b/src/sc.h
@@ -99,6 +99,12 @@
 /* include system headers */
 
 #include <math.h>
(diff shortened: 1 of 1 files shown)


gghosh8495:
All values for constants have been copied from the visual C++ docs.
https://docs.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170

PR #69, opened 2022-03-09
Use autoconf's `ac_c_werror_flag` for compiler-portable qsort_r compatibility check

tisaac:
The `#pragma GCC diagnostic error` approach does not work when using,
for example, the intel compiler.

### DIFF
--- a/config/sc_qsort.m4
+++ b/config/sc_qsort.m4
@@ -10,6 +10,16 @@ dnl The PREFIX argument is currently unused but should be supplied.
 dnl
 AC_DEFUN([SC_CHECK_QSORT_R], [
 
+   AC_LANG_PUSH([C])
+
+   dnl We want any warning about missing declarations or incompatible pointer
+   dnl types to trigger an error: in a belt-and-suspenders approach, we have
+   dnl the `#pragma GCC diagnostic error` statements in the source, but we also
+   dnl uses this compiler's version of "-Werror" that autoconf already knows
+   dnl about for this test
+   sc_check_qsort_r_save_werror_flag=$ac_c_werror_flag
+   ac_c_werror_flag=yes
+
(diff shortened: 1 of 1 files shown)


gghosh8495:
@cburstedde I think we should add a workflow file for Intel compilers, what do you think ?

cburstedde:
> @cburstedde I think we should add a workflow file for Intel compilers, what do you think ?

Yes, that would be a good thing.

cburstedde:
The portable -Werror turns pragma warnings into errors, too. Then if #pragma GCC is not recognized, we fail.
I've tentatively deactivated this flag. If we reactivate, we'd have to remove the pragmas altogether -- which is better?

gghosh8495:
I think the -Werror is not invoked inside the qsort_r link test. If I remember correctly, I added those diagnostic pragma's due to that.

gghosh8495:
I looked up the Intel c compiler man page, I did not find an equivalent pragma. But then again I am not fully aware of Intel compilers, I will search more and hopefully stumble upon something useful.

jedbrown:
In PETSc and libCEED, we use `_Pragma("GCC ivdep")` or whatever, which can be placed in a macro (unlike `#pragma`). `_Pragma` is part of C99.

Note that the new Intel compilers (`icx`) are based on LLVM and thus behave similarly to upstream clang.

cburstedde:
> In PETSc and libCEED, we use `_Pragma("GCC ivdep")` or whatever, which can be placed in a macro (unlike `#pragma`). `_Pragma` is part of C99.

Ok, I'll stay away from them then. :) Trying to be as oldschool as possible.

> Note that the new Intel compilers (`icx`) are based on LLVM and thus behave similarly to upstream clang.

Cool that may mean thet acceptance of GCC pragmas may widen.

cburstedde:
Ok, misclicked. Merging in the current form.

PR #71, opened 2022-05-06
refactor cmakedefine

gghosh8495:
# Title Refactor cmake build

Following up on issue # .

Proposed changes:
set #cmakedefines to 1

### DIFF
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -75,7 +75,6 @@ endif()
 check_symbol_exists(backtrace execinfo.h SC_HAVE_BACKTRACE)
 check_symbol_exists(backtrace_symbols execinfo.h SC_HAVE_BACKTRACE_SYMBOLS)
 
-# check_include_file(dlfcn.h SC_HAVE_DLFCN_H)
 check_include_file(execinfo.h SC_HAVE_EXECINFO_H)
 check_symbol_exists(fsync unistd.h SC_HAVE_FSYNC)
 check_include_file(inttypes.h SC_HAVE_INTTYPES_H)
--- a/cmake/git.cmake
+++ b/cmake/git.cmake
@@ -1,10 +1,13 @@
 # --- extract version from Git
 
+set(PROJECT_MAJOR 0)
+set(PROJECT_MINOR 0)
+set(PROJECT_PATCH 0)
 set(PROJECT_VERSION 0.0.0)
 find_program(GIT_VERSION_GEN NAMES git-version-gen
              PATHS ${CMAKE_SOURCE_DIR}/build-aux NO_DEFAULT_PATH)
 if(GIT_VERSION_GEN)
-  execute_process(COMMAND ${GIT_VERSION_GEN} .tarball_version
+  execute_process(COMMAND ${GIT_VERSION_GEN} .tarball-version
     WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     RESULT_VARIABLE _err
     OUTPUT_VARIABLE git_version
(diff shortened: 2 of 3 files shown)


gghosh8495:
@cburstedde I have checked the C macros and they are set correctly as far as I see. I don't see any reason for any odd outcome. The CI on git does not seem to complain. I have removed some macros which were not set by config.cmake

cburstedde:
Thanks! I can see two points remaining:

 1. There is the cmakedefine to `@SC_ENABLE_DEBUG@`, which seems fine to me. `SC_ENABLE_MPI`, on the other hand, is set without any variable. Are there reasons for/against these two to be unified, basically referring `@SC_..._MPI@` variables?
 1. For the source version numbers, p4est uses `#define P4EST_VERSION_MAJOR "@PROJECT_MAJOR@"`. Can we add the analogous functionality to libsc?

gghosh8495:
Ah I see what you are asking. In this context it does not really matter if you use 1 or @SC_ENABLE_MPI@ because it always set to some value in config.cmake or cmakelist.txt
But this raises a question that why set every #cmakedefine to 1 because #cmakedefine is only a if-else for #define and #undef and if for some reason a macro is 0 in config.cmake/cmakelist.txt then we set the wrong value.
I would like to ask Bruno about this.

gghosh8495:
@cburstedde The version numbering has been integrated. As for SC_ENABLE_DEBUG macro I read all the files and played around with it but could not find a solid reason to use the @SC_ENABLE_DEBUG@ construction. Either way  the tests do not complain.

gghosh8495:
Let me know how would you like to keep it ?

cburstedde:
On Sun, Jun 05, 2022 at 02:06:59PM -0700, gghosh8495 wrote:
> @cburstedde The version numbering has been integrated. As for SC_ENABLE_DEBUG macro I read all the files and played around with it but could not find a solid reason to use the @SC_ENABLE_DEBUG@ construction. Either way  the tests do not complain.

Thanks, then just changing it to cmakedefine 1 would be the general solution?

gghosh8495:
@cburstedde I fixed all issues I found yesterday. Its a very simple fix, for SC_VERSION_MAJOR, SC_VERSION_MINOR and SC_VERSION_PATCH we assign an int and not a char* . This same mistake has been repeated by Bruno in his p4est cmake commits. 

Everything checks out on my work machine and git CI. This PR is ready to be merged.

As for README, libsc already has cmake build instructions.

cburstedde:
> @cburstedde I fixed all issues I found yesterday. Its a very simple fix, for SC_VERSION_MAJOR, SC_VERSION_MINOR and SC_VERSION_PATCH we assign an int and not a char* . This same mistake has been repeated by Bruno in his p4est cmake commits. 

It is quite possible that at least the patch version may contain letters and
other characters.  Is there no way to preserve the generality of the char*?

gghosh8495:
We can let SC_VERSION_PATCH be a char* but the other two have to be integers because of this code 

int
p4est_version_major (void)
{
  /* In rare cases SC_VERSION_MINOR may be a non-numerical string */
  return sc_atoi (SC_TOSTRING (P4EST_VERSION_MAJOR));
}

cburstedde:
> We can let SC_VERSION_PATCH be a char* but the other two have to be integers because of this code 

They don't "have to."  This code works fine for a string like 5-abcd.

The actual think to ensure may rather be that the SC_VERSION_ variables are set
to the correct contents based on the git describe output.

> int
> p4est_version_major (void)
> {
>   /* In rare cases SC_VERSION_MINOR may be a non-numerical string */
>   return sc_atoi (SC_TOSTRING (P4EST_VERSION_MAJOR));
> }

cburstedde:
Thanks, please proceed with minimizing the comment diff where it makes sense and double-check it one more time.

gghosh8495:
@cburstedde I have made the necessary changes. Can you please take a look ?

cburstedde:
Thanks, just in case, would it make sense to minimize the diff between p4est and sc wrt. the config.h.in and config.cmake?

PR #72, opened 2022-06-17
Prepare array_is_sorted for const qvt

mkirilin:
Make it possible to pass a constant quadrant vtable to the function sc3_array_is_sorted.


mkirilin:
I updated the p4est3 part adding a wrapper for the passing qvt. No modifications into sc3 are needed. Please, see the pull request on BB.

PR #73, opened 2022-07-04
Feature tarball test

gghosh8495:
# Title

Following up on issue # .

Proposed changes:


gghosh8495:
@cburstedde I think we need to add .git files into the cmake tarball because the dist-hook on autotools requires git-version-gen

dist-hook:
	echo $(VERSION) > $(distdir)/.tarball-version
	(GITGEN_VERSION=`(cd @top_srcdir@ && ./build-aux/git-version-gen\
                .tarball-version)` ; \
	 test "x$(VERSION)" = "x$${GITGEN_VERSION}" || \
         ((echo "Stale version;"; echo $(VERSION); echo "$${GITGEN_VERSION}"; \
           echo "Please run:" ; \
           echo "     (cd @top_srcdir@ && ./bootstrap)" ; \
           echo "before make dist") 1>&2 ; rm -r $(distdir) ; exit 1))

What do you think ?

gghosh8495:
@cburstedde The .tarball-version file only exists when we unpack a tarball, in our case we do a fresh git checkout and then create a tarball. This written in the documentation of git-version-gen. So, in case of autotools generating .tarball-version file will not be possible, right ?

PR #74, opened 2022-07-05
CMake enhancements

scivision:
* CI: unify techniques with p4est
* test: simplify, better approach for Windows

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
       - "**/CMakeLists.txt"
       - "**.cmake"
       - "**.cmake.in"
-      - ".github/workflows/ci_cmake.yml"
+      - ".github/workflows/ci.yml"
   pull_request:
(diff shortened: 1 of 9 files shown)


gghosh8495:
@cburstedde It seems qsort_r may not be the issue because as per the configure script output it is able to detect qsort_r in BSD format. That is what we would expect on a mac. I wish there was a way to look at error log on git actions.

scivision:
there is a way to have Github Actions auto-upload the config.log on error. Another way is to just print it like:

```yaml
    - name: print config log
      if: ${{ failure() }}
      run: cat build/CMakeFiles/CMakeError.log
 ```

scivision:
the config error is a coincidence -- the Github Actions .yml file was mistakenly set to run when ci_cmake.yml changed instead of ci_darwin.yml

cburstedde:
I see, thanks. Manual merging.

PR #75, opened 2022-07-25
cmake: sc_config.h escape spaces

scivision:
cmake needs to have spaces escaped for generating sc_config.h so that CMake doesn't insert semicolons, which are CMake's separator for lists. Gautam had noticed this in the CMake-generated sc_config.h

As a separate matter, used `set_property()` in CMake, which is a more robust syntax than `set_tests_properties` and `set_target_properties` etc.


cburstedde:
Thanks!

PR #76, opened 2022-07-26
Fix example build

gghosh8495:
Make libsc cmake build example directory

Proposed changes:
1. Add examples directory to top level CMakeLists.txt
2. Drop the stringent requirement to find libsc within libsc
3. Avoid treating examples as tests


CMakeLists.txt
```diff
@@ -67,6 +67,7 @@ add_subdirectory(src)
 
 if(BUILD_TESTING)
   add_subdirectory(test)
+  add_subdirectory(example)
```
cburstedde, line 70: Please always build the examples, not only when testing.

example/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ find_package(Threads)
 
 # --- find our library
 
-find_package(SC REQUIRED)
+find_package(SC QUIET)
```
cburstedde, line 25: If we don't expect to find this, would everything work by removing this line entirely?

CMakeLists.txt
gghosh8495: Done!

PR #80, opened 2022-08-02
cmake: allow "example" to be used by itself or under main project

scivision:


scivision:
@gghosh8495 this by default doesn't run example tests but builds the examples if BUILD_TESTING. I think your other PR builds examples by default but also removes the ability to test them.

cburstedde:
> @gghosh8495 this by default doesn't run example tests but builds the examples if BUILD_TESTING. I think your other PR builds examples by default but also removes the ability to test them.

Yes, how we do it with autotools is to always build the examples but to never run them as tests.
With autotools, we build and run the tests on make check and only then.
If we could do it the same way with CMake, that would be great.
This applies to both libsc and p4est.

cburstedde:
> > @gghosh8495 this by default doesn't run example tests but builds the examples if BUILD_TESTING. I think your other PR builds examples by default but also removes the ability to test them.
> 
> Yes, how we do it with autotools is to always build the examples but to never run them as tests. With autotools, we build and run the tests on make check and only then. If we could do it the same way with CMake, that would be great. This applies to both libsc and p4est.

It would be totally fine, with CMake, to always compile the tests with the build and only run them on ctesting.

cburstedde:
Thanks, let's go with this and possibly tune some more later.

PR #84, opened 2022-08-25
cmake:package/install: flatten CMake meta, correct CPack parameters

scivision:
Previously, CPack parameters were incorrectly set that disabled
packaging from
a superproject as used from gemini3d/external to make a project-wide
archive from a single command over all packages.

Install of CMake metadata is flattened per modern practice


cburstedde:
With recent merges there is a conflict that's likely trivial to address for @scivision.

P. S. Slight preference to use custom branch names instead of pushing to prev3-develop, thanks again!

cburstedde:
Thanks! I've merged in the latest prev3-develop branch and hopefully resolved the conflict.
What keeps bugging me, however, are the failures of the CMake CI test matrix. Is there a way to make them go away reliably? I like the general idea but would like it best under robust operation.

cburstedde:
I've tried rerunning the CI several times and error probability remains high. I have one suggestion though which might help: With the autotools, we are no longer compiling any openmp code. This means it has disappeared from src and we are no longer building `example/openmp`. If you like to deactivate any openmp test code and set the configuration to permanently false, this would be ok.

cburstedde:
The ctest error on Windows seems quite reproducible (error code 0xc0000135 for five tests).

cburstedde:
Ok, it seems we're getting there. Not running examples as tests on Windows seems to help, and to switch from OpenMPI to MPICH on Linux as well.

PR #86, opened 2022-09-12
update for conforming with p4est darwin ci

gghosh8495:


cburstedde:
It seems this changes the compiler from C++ back to C. This is not what we intended.
The original problem is to have the C++ compiler defined by mpicxx to find the zlib function `adler32_combine` during `configure`, so how can this be accomplished?

cburstedde:
From the `config.log file` it looks like the test program for `adler32_combine` calls it with a prototype that does not match the zlib definition. This usually works in C, but apparently not in C++. Would we want a version of `SC_CHECK_LIB` that accepts a prototype as input and produces a working test program?

PR #89, opened 2022-09-30
Initialize output of sc_{strcopy,snprintf}

tim-griesbach:
# Initialize output of sc_{strcopy,snprintf}

Proposed changes: We initialize the output string of `sc_{strcopy,snprintf}` with `'\0'`. Therefore, the output of the mentioned function is now initialized even after the first `'\0'`. The initialization is equivalent to padding the output to the right with `'\0'`.

### DIFF
--- a/src/sc.c
+++ b/src/sc.c
@@ -1520,6 +1520,9 @@ sc_snprintf (char *str, size_t size, const char *fmt, ...)
     return;
   }
 
+  /* avoid uninitialized bytes if strlen (src) < size - 1 */
+  memset (str, (int) '\0', size);
+
   /* Writing this function just to catch the return value.
      Avoiding -Wnoformat-truncation gcc option this way */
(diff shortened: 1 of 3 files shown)


src/sc.c
```diff
@@ -1520,6 +1520,9 @@ sc_snprintf (char *str, size_t size, const char *fmt, ...)
     return;
   }
 
+  /* avoid uninitialized bytes if strlen (src) < size - 1 */
+  memset (str, (int) '\0', size);
```
cburstedde, line 1524: There is no cast needed, a simple 0 (just the number, no backslash) is best with memset (in general as well).

tim-griesbach: Okay, I removed the cast and set the parameter to `0`.

cburstedde: Thanks. I removed such casts in three other places as well.

PR #92, opened 2022-11-16
Feature encode

cburstedde:
# Generic compressed ASCII encode and decode

Proposed changes: Add functions to compress and encode arbitrary data into ASCII.
The proposed implementation relies on zlib for compression and base64 for encoding.
If zlib is not detected by configure, we emulate non-compressed zlib operation.

The test_helpers program has been extended to exercise the new functions.

### DIFF
--- a/src/sc_io.c
+++ b/src/sc_io.c
@@ -396,17 +396,21 @@ sc_io_encode (sc_array_t *data, sc_array_t *out)
   size_t input_size;
   size_t base64_lines;
   size_t encoded_size;
-  size_t zlin, irem, ocnt;
+  size_t zlin, irem;
+#ifdef SC_ENABLE_DEBUG
+  size_t ocnt;
+#endif
   sc_array_t compressed;
   base64_encodestate bstate;
 
   SC_ASSERT (data != NULL);
   if (out == NULL) {
-    /* in-place operation */
+    /* in-place operation on string */
     SC_ASSERT (SC_ARRAY_IS_OWNER (data));
+    SC_ASSERT (data->elem_size == 1);
   }
   else {
-    /* data is placed in output array */
+    /* data is placed in output string */
     SC_ASSERT (SC_ARRAY_IS_OWNER (out));
     SC_ASSERT (out->elem_size == 1);
   }
@@ -430,11 +434,9 @@ sc_io_encode (sc_array_t *data, sc_array_t *out)
 
   /* prepare output array */
   if (out == NULL) {
-    /* to do: write function sc_array_reshape */
-    data->elem_size = 1;
-    data->elem_count = input_size;
     out = data;
   }
+  SC_ASSERT (out->elem_size == 1);
   input_size = (size_t) input_compress_bound;
   base64_lines = (input_size + 53) / 54;
   encoded_size = 4 * ((input_size + 2) / 3) + base64_lines + 1;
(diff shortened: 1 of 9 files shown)


src/sc_io.h
```diff
@@ -257,6 +257,164 @@ int                 sc_io_source_read_mirror (sc_io_source_t * source,
                                               size_t bytes_avail,
                                               size_t *bytes_out);
 
+/** Return a boolean indicating whether zlib has been configured.
+ * \return          True if zlib has been found on runnning configure,
```
tim-griesbach, line 261: There is a typo runnning -> running.

src/sc_io.c
```diff
@@ -384,6 +385,622 @@ sc_io_source_read_mirror (sc_io_source_t * source, void *data,
...
+#if SC_IO_DBC % 3 != 0
+#error "SC_IO_DBC must be a multiple of 3"
+#endif
+
+/* byte count for one line of base64 encoded data and newline */
+#define SC_IO_LBC (SC_IO_DBC / 3 * 4)
+#define SC_IO_LBD (SC_IO_LBC + 1)
+#define SC_IO_LBE (SC_IO_LBC + 2)
+
+/* see RFC 1950 and RFC 1950 for the uncompressed zlib format */
```
tim-griesbach, line 409: A typo: "RFC 1950 and RFC 1950".

tim-griesbach (approved):
This looks good to me. I just commented on two typos.

src/sc_io.h
cburstedde: Thanks!

tim-griesbach (commented):
I added one comment on the scspell change.

scspell
```diff
@@ -2,7 +2,7 @@
 
 CSP=codespell
 IGN=doc/codespell_ignore.txt
-CSO="-I $IGN $@"
+CSO="-x $IGN $@"
```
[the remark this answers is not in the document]
cburstedde: Thanks for checking, I'm dropping the change from this branch. (Feel free to resolve the conversation.)

cburstedde:
I've added an argument for future soft error reporting to the decode prototypes.
The encode prototypes do not seem to need it, the question is whether we should add it just in case.
This would then require the calling code to handle this error, which creates a bit of a hassle.

tim-griesbach:
> I've added an argument for future soft error reporting to the decode prototypes.
> The encode prototypes do not seem to need it, the question is whether we should add it just in case.
> This would then require the calling code to handle this error, which creates a bit of a hassle.

It seems to me that the only possible need for soft errors for the decode functions would be for `sc_io_encode_zlib` an invalid compression level. Just as a comparison: MPI I/O gives a soft error if the access mode of a file is invalid. However, if we do not want to do something like that, I agree we do not need such a soft error object.

cburstedde:
> > I've added an argument for future soft error reporting to the decode prototypes.
> > The encode prototypes do not seem to need it, the question is whether we should add it just in case.
> > This would then require the calling code to handle this error, which creates a bit of a hassle.
> 
> It seems to me that the one possible need for soft errors for the decode functions would be for `sc_io_encode_zlib` an invalid compression level. Just as a comparison: MPI I/O gives a soft error if the access mode for a file is invalid. However, if we do not want to do something like that, I agree we do not seem to need such a soft error object.

This is a good point. I had thought about what users would do without zlib -- they wouldn't know what levels to pass in at all. What about defining an enum for sc's best, average, and no compression that does not depend on zlib? Then we can hard assert on its range of values.

PR #94, opened 2022-11-21
Feature sc3 mpienv

mkirilin:
Add some auxiliary functions of getting data from a mpi environment object.


mkirilin:
Added getting the total number of nodes by mpienv structure. Also added comments-warnings for getting c-arrays of nodes sizes and nodes offsets.

src/sc3_mpienv.c
```diff
@@ -423,3 +423,53 @@ sc3_mpienv_get_node_frank (const sc3_mpienv_t * m, int *node_frank)
...
+{
+  SC3E_RETVAL (num_nodes, -1);
+  SC3A_IS (sc3_mpienv_is_setup, m);
+
+  *num_nodes = m->num_nodes;
+  return NULL;
+}
+
+sc3_error_t        *
+sc3_mpienv_get_node_sizes (const sc3_mpienv_t * m, int **node_sizes)
```
cburstedde, line 448: Thanks! What would be your opininon about trying the `const int **` type?

mkirilin: It makes sense to add it for safety reasons.

cburstedde (approved):

PR #97, opened 2022-12-05
CMake: build Zlib instead of find for compatibility

### LINKED ISSUE #96: CMake: build Zlib instead of find
## Description

CMake-based users of libsc (and higher-level libraries using libsc) have noticed issues with ABI compatibility of Zlib. This is not a CMake-specific issue, but to resolve this we are creating a PR that when `-Dzlib=true` (default true) Zlib is automatically downloaded and built first.

scivision:
fixes #96 

Proposed changes: for CMake, build Zlib instead of finding to avoid link-time build errors.

For CI, the default is about to change to Ubuntu-22.04. To keep coverage of older compilers, lock the Linux image to Ubuntu 22.04 and use the pre-installed compilers to save time.

The libsc CMake package is also enhanced to publish information about its capabilities to avoid libsc-using project to have to duplicate the same capability checks already known by libsc, and that need to be consistent with libsc as-built.


cburstedde:
Thanks!

I'm debating to prepare two #defines per library, one for the header like SC_HAVE_ZLIB_H, and one for a working compile-link-test, like SC_HAVE_ZLIB. I'd try to establish this default for all third-party libraries. Would you like to provide the second one for ZLIB, or would you rather think this unnecessary and we should keep to one #define?

In addition, it will be good to check for/enable `SC_ENABLE_MPIIO`, which refers to the MPI_File functions. Without this flag, parallel I/O falls back to something rather horribly slow. Ideally, when MPI is on, MPI I/O should be on.

I can merge as is, or wait for updates to this PR.

scivision:
I put a question in #96 if you want to make "-Dzlib" a 3-way option: off, find, or build, with build being default

cburstedde:
Thanks, I've commented in the issue. I'm also thinking to simplify the defines: just one per library, so SC_HAVE_ZLIB is fine and defined when the compile/link test works that is implicitly including zlib.h, and SC_HAVE_JSON is defined when jansson.h is found and the link test finds json_integer in -ljansson.

PR #99, opened 2022-12-16
Feature sc3 cut

mkirilin:
A small update of lib-sc3:

1. Add a 64-bit integer cut function.
2. Add an assembling check on availability of socket communicator split for MPI3 shared memory. If so, split comm by socket always.

### DIFF
--- a/config/sc_mpi.m4
+++ b/config/sc_mpi.m4
@@ -23,7 +23,7 @@ dnl If MPI is enabled, set AC_DEFINE and AC_CONDITIONAL for PREFIX_ENABLE_MPI.
 dnl If MPI I/O is not disabled, set these for PREFIX_ENABLE_MPIIO.
 dnl If MPI_Init_thread is not disabled, set these for PREFIX_ENABLE_MPITHREAD.
 dnl If MPI shared nodes are supported, set these for PREFIX_ENABLE_MPISHARED.
-dnl If MPI shared memory is continious, set these for PREFIX_ENABLE_MPISOCKET.
+dnl If MPI split by socket is supported, set these for PREFIX_ENABLE_MPISOCKET.
 dnl
 dnl SC_MPI_ENGAGE(PREFIX)
 dnl
--- a/config/sc_mpi.m4
+++ b/config/sc_mpi.m4
@@ -431,6 +431,28 @@ MPI_Finalize ();
  $2])
 ])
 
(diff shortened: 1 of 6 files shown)


cburstedde:
Thanks; is it possible to fix the CI failures?

mkirilin:
> Thanks; is it possible to fix the CI failures?
> It may help to merge the latest prev3-develop into feature-sc3.

Issues with clang are fixed by @tim-griesbach 's [pull request ](https://github.com/cburstedde/libsc/pull/98). Once it's merged, I update feature-sc3 branch with the resent prev3-develop changes.

Issues with valgrind are caused by MPI implementation, somewhere inside the function allocating shared memory.

cburstedde:
@tim-griesbach 's changes are still in the diff while they should not be. Is it possible to make github pick up on the latest merges and pushes of feature-sc3?

mkirilin:
> @tim-griesbach 's changes are still in the diff while they should not be. Is it possible to make github pick up on the latest merges and pushes of feature-sc3?

Done. Undid merge of Tim's branch to the current feature-sc3-cut branch. Merged an actual feature-sc3 into feature-sc3-cut instead.

config/sc_mpi.m4
```diff
@@ -23,6 +23,7 @@ dnl If MPI is enabled, set AC_DEFINE and AC_CONDITIONAL for PREFIX_ENABLE_MPI.
 dnl If MPI I/O is not disabled, set these for PREFIX_ENABLE_MPIIO.
 dnl If MPI_Init_thread is not disabled, set these for PREFIX_ENABLE_MPITHREAD.
 dnl If MPI shared nodes are supported, set these for PREFIX_ENABLE_MPISHARED.
+dnl If MPI shared memory is continious, set these for PREFIX_ENABLE_MPISOCKET.
```
cburstedde, line 26: Please spell-check, also do you mean continuous or contiguous, and how does it relate to TYPE?

I was under the impression that the setting of contiguous shared memory is an independent quality from TYPE_SHARED and TYPE_SOCKET.

mkirilin: Updated the description.

mkirilin:
I updated the PR with the latest prev3-develop changes and kept the non-contiguous shared memory tests secret from Valgrind. Now all the CI tests pass.

PR #102, opened 2022-12-20
Feature sc3

mkirilin:
Merge the recent prev3-develop changes.
It's the best to merge it before [feature-sc3-cut](https://github.com/cburstedde/libsc/pull/99) PR to reduce the diff.


mkirilin:
CI fails, because the fix is on [feature-sc3-cut](https://github.com/cburstedde/libsc/pull/99) PR.

PR #108, opened 2023-01-12
Fix sc3 warnings

mkirilin:
1. Fixed some warnings appeared with the older versions of gcc.
2. Check availability of shared memory and socket split properly for cmake build case.


cburstedde:
Thanks! Let's keep an eye out for warnings in the future, checking various compilers, with/without MPI, with/without debug, CMake/autoconf on some regular schedule.

Issue #107, opened 2023-01-02
Non-MPI build unconditionally depends on `ompi` and `orte` pkgconfig 


StefanBruens:
https://github.com/cburstedde/libsc/blob/1e64da6919f5fbba423e2fcac102c51e6c0cad49/cmake/pkgconf.cmake#L2

StefanBruens:
Seems ok so far.

PR #110, opened 2023-01-19
CMake: correct pkgconf generation

scivision:
# CMake:pkgconf: only specify used requirements

Following up on issue #107 .


StefanBruens:
`orte` should be omitted, it is a runtime, not a build time dependency.

PR #114, opened 2023-04-06
Feature make install

cburstedde:
# Extend --with-sc= configure logic

## Starting point
This PR is about using libsc as subpackage as well as {other packages that use libsc} as subpackage.
Previously, we could set `--with-sc=<path to installation>` to specify the toplevel directory of an installed library, but we would still expect `libsc.la` to be present. But usually, this file should not be relied upon from the outside.

## Proposed changes:
1. The behaviour of `--with-sc=<path>` is changed to use the usual `-L` and `-l` flags to the library.
2. We add `--with-sc=external` as a literal option to not pull in any compile, link and runtime flags at all. In this case, the package builder must ensure that LDFLAGS, CPPFLAGS are set properly, and LD_LIBRARY_PATH to run any examples.

The upstream `p4est` and `t8code` libraries will not run with this libsc update unless submodules are updated.


cburstedde:
> This looks good to us. We noticed that before the changes of your PR it was possible to use a relative path for `--with-sc=<path>`. However, after applying the changes of the PR we obtain
> 
> ```
> libtool:   error: only absolute run-paths are allowed
> ```
> 
> as an error message for running `make check` after configuring `--with-sc=<path>` with a relative path and without `-l` and `-L` but with `-I`. Namely, we have used in this case
> 
> ```
> ./configure --with-sc="<RELATIVE_PATH>/local/" --enable-mpi CFLAGS="-Wall -g -O0" CPPFLAGS="-I<RELATIVE_PATH>/local/include"

Thanks for checking; does the error come from the -I part or the --with-sc part?

tim-griesbach:
> Thanks for checking; does the error come from the -I part or the --with-sc part? (I'd suspect the latter.)

We tested the code with the changes of your PR with the configure command
```
./configure --with-sc="<RELATIVE_PATH>/local/" --enable-mpi CFLAGS="-Wall -g -O0" CPPFLAGS="-I<ABSOLUTE_PATH>/local/include"
```
for p4est. Then, `make check` does not work with the same error message as above. Therefore, `--with-sc` seems to cause the error message.

> It wouldn't be too bad if --with-sc requires an absolute path due to using it with -L. --prefix has the same requirement.

Maybe users still use `--with-sc=<path>` without `-L`. However, the error message is rather instructive.
Furthermore, we tested
```
./configure --with-sc=external --enable-mpi CFLAGS="-Wall -g -O0" LDFLAGS="-L<RELATIVE_PATH>/local/lib" LIBS="-lsc" CPPFLAGS="-I<RELATIVE_PATH>/local/include"
```
as configure command in p4est on branch feature-make-install. So `-L` worked for us with a relative path.

cburstedde:
Thanks for pointing out the necessity for CPPFLAGS by the way. I've removed the need to specify it alongside --with-sc. Does it work omitting it explicitly for a --with-sc=<absolute path>?

cburstedde:
I believe the error message with a relative path is due to specifying -rpath in the Makefile.am. Without -rpath, the examples and tests do not find the installed library. Is your experience consistent with this?

cburstedde:
> I believe the error message with a relative path is due to specifying -rpath in the Makefile.am. Without -rpath, the examples and tests do not find the installed library. Is your experience consistent with this?

It may be possible to add some script magic to absolutify the path given to `$1_$1_RPATH`. Worth it?

tim-griesbach:
> Thanks for pointing out the necessity of CPPFLAGS by the way. I've removed the need to specify it alongside --with-sc. Does it work omitting it explicitly for a `--with-sc=<absolute path>`?

Yes,
```
./configure --with-sc="<ABSOLUTE_PATH>/local" --enable-mpi CFLAGS="-Wall -g -O0"
```
works now.


> I believe the error message with a relative path is due to specifying -rpath in the Makefile.am. Without `-rpath`, the examples and tests do not find the installed library. Is your experience consistent with this?

Yes, without `-rpath` I get
```
gcc: error: P4EST_SC_RPATH@: No such file or directory
```
as error message for `make check` in p4est.



> > I believe the error message with a relative path is due to specifying -rpath in the Makefile.am. Without -rpath, the examples and tests do not find the installed library. Is your experience consistent with this?
> 
> It may be possible to add some script magic to absolutify the path given to `$1_$1_RPATH`. Worth it?

We could use the POSIX commands `dirname`, `pwd`, `basename` to get the absolute path without too much effort but we would must check carefully corner cases what did not look easy at the first glance.

cburstedde:
> Yes, without `-rpath` I get
> 
> ```
> gcc: error: P4EST_SC_RPATH@: No such file or directory
> ```
> 
> as error message for `make check` in p4est.

This is a bit weird, as if the @P4EST_SC_RPATH@ substitution gets ripped to pieces.

> > It may be possible to add some script magic to absolutify the path given to `$1_$1_RPATH`. Worth it?
> 
> We could use the POSIX commands `dirname`, `pwd`, `basename` to get the absolute path without too much effort but we would must check carefully corner cases what did not look easy at the first glance.

I agree, let's not go there.

tim-griesbach (approved):
From our side this looks good and as we can leave it as it is.

cburstedde:
Thanks also to @jmark for mentioning library-library dependencies.

Added a package-specific LIBADD variable to be used from p4est etc.

jmark:
I tested this PR together with https://github.com/DLR-AMR/t8code/pull/506 and https://github.com/DLR-AMR/t8code/pull/506 and everything works as expected.

From my side this PR can be merged as soon as possible.

@cburstedde @sloede

jmark:
With this PR I could finally register an official [t8code_jll](https://juliahub.com/ui/Packages/t8code_jll/6PmXJ/1.1.2+2) Julia package. But this is only a temporary solution, since this builds on pre-releases and branches of sc, p4est and t8code. If this PR will be merged in the near future, the package could be build upon official releases of the three libraries.

sloede:
> With this PR I could finally register an official [t8code_jll](https://juliahub.com/ui/Packages/t8code_jll/6PmXJ/1.1.2+2) Julia package.

Great news 🎉 , congratulations!

Just so I understand: The next step would be to merge this PR, create a new official release of p4est, then create a new official release of t8code based on the new p4est release, then update the p4est and t8code JLL packages?

jmark:
> > With this PR I could finally register an official [t8code_jll](https://juliahub.com/ui/Packages/t8code_jll/6PmXJ/1.1.2+2) Julia package.
> 
> Great news 🎉 , congratulations!
> 
> Just so I understand: The next step would be to merge this PR, create a new official release of p4est, then create a new official release of t8code based on the new p4est release, then update the p4est and t8code JLL packages?

Yes! :)

jmark:
@tim-griesbach @cburstedde Any progress on this PR?

PR #117, opened 2023-04-14
Do not require compile-time MPI constants for configure

sloede:
The MPI standard does not mandate constants to be compile-time constants, but only to be link-time constants. In the current setup for configure, however, MPI constants are used in a C `switch` statement, for which they need to be compile-time constants.

This PR tries to remedy this by using `if` statements instead of switches.

@jmark Can you please check if this patch fixes your issue in  https://github.com/cburstedde/p4est/issues/197?

### DIFF
--- a/config/sc_mpi.m4
+++ b/config/sc_mpi.m4
@@ -149,7 +149,7 @@ if test "x$HAVE_PKG_MPI" = xyes ; then
   elif test "x$$1_MPIRUN" = xmpiexec ; then
     $1_MPI_TEST_FLAGS="-n 2"
   else
-    AC_MSG_ERROR([--enable-mpi given but neither mpirun nor mpiexec found])
+    $1_MPIRUN=
   fi
 fi
 dnl We are not using precious variables for simplicity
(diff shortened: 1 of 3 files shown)


jmark:
@sloede @cburstedde 

The change from case- (https://github.com/cburstedde/libsc/commit/cfc2ca3cde30950d57287d3c63f0d062afc3632e)
to if-statements did the trick.  

Also, I added a workaround for the error
```
--enable-mpi given but neither mpirun nor mpiexec found
```
on the `x86_64-w64-mingw32-mpi+microsoftmpi` platform.

Basically reverting to the original state before this error message was added in
https://github.com/cburstedde/libsc/commit/483980fe725ad3ea5a68a54cfd5ef9f405071267.

config/sc_config_autotools.h.in~
sloede, line 1: @jmark Did you really want to commit this file?

jmark: Sorry this slipped in. Gonna remove it.

cburstedde:
Thanks all! Good catch. If you would please rebase such that the accidental file additions disappear?

cburstedde:
> > Thanks all! Good catch. If you would please rebase such that the accidental file additions disappear?
> 
> Hope I've done that right.
> 
> @sloede Commits should be ready to merge.

Awesome; may I also ask the two of you to add the usual author files under doc?

sloede (commented):
LGTM and is imho ready for a review by @cburstedde 🙂

sloede:
> Awesome; may I also ask the two of you to add the usual author files under doc?

Done, both author files have been added.

cburstedde:
Thanks so much; I appreciate your working together on this.

PR #118, opened 2023-04-17
Update sc_io.c

pkestene:
Fix build when using cmake build, using cmake option zlib=OFF. 
See  recent p4est issue : https://github.com/cburstedde/p4est/issues/198 where the problem was noticed

### DIFF
--- a/src/sc_io.c
+++ b/src/sc_io.c
@@ -26,6 +26,11 @@
 #include <libb64.h>
 #ifdef SC_HAVE_ZLIB
 #include <zlib.h>
+#else
+/* when SC_HAVE_ZLIB is undefined, we need to define Z_BEST_COMPRESSION (zlib.h compatibility) */
+#ifndef Z_BEST_COMPRESSION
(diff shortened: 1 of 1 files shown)


src/sc_io.c
```diff
@@ -647,6 +647,12 @@ sc_io_nonuncompress (char *dest, size_t dest_size,
 
 #define SC_IO_ENCODE_INFO_LEN 9
 
+/* when SC_HAVE_ZLIB is undefined, we need to define Z_BEST_COMPRESSION (zlib.h compatibility) */
+#ifndef Z_BEST_COMPRESSION
+#define Z_BEST_COMPRESSION 9
```
cburstedde, line 652: Thanks! What would you think about moving these three lines to the very top into the `#else` branch of
`#ifdef SC_HAVE_ZLIB`? Then we would survive if Z_BEST_COMPRESSION becomes a variable in zlib.

pkestene: agreed, it's better.

PR #119, opened 2023-04-18
Add json support in cmake build system.

pkestene:
# Add json support in cmake build

### DIFF
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -60,6 +60,9 @@ test_sc_example(logging logging/logging.c)
 test_sc_example(test_shmem testing/sc_test_shmem.c)
 
 if(SC_HAVE_GETOPT_H)
+  configure_file(options/sc_options_example.ini sc_options_example.ini COPYONLY)
+  configure_file(options/sc_options_example.json sc_options_example.json COPYONLY)
+  configure_file(options/sc_options_preload.ini sc_options_preload.ini COPYONLY)
   test_sc_example(options options/options.c)
 endif()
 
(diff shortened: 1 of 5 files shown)


cburstedde:
Thanks; this is really helpful. It seems it works with Windows and all? :)

example/CMakeLists.txt
```diff
@@ -60,6 +60,9 @@ test_sc_example(logging logging/logging.c)
 test_sc_example(test_shmem testing/sc_test_shmem.c)
 
 if(SC_HAVE_GETOPT_H)
```
cburstedde, line 62: I'm wondering whether we need to check for this #define.
If it is not defined, we are building a replacement -- options should always work the same.

pkestene: In principle you are right, but currently in that regard the autotools and cmake build system are not equivalent:

- in the autotools build, there is some logic to define SC_PROVIDE_GETOPT symbol
- this logic does not exist in the cmake build (SC_PROVIDE_GETOPT is not controlled in cmake); so currently the cmake build only check for header getopt.h and use that info to decide if or if sc_options.c, etc are included in the build

maybe we could add the functionnality in the cmake build (in another MR).

cburstedde: > Currently in cmake build, SC_PROVIDE_GETOPT is not defined, so only getopt.h can be used, that way sc_options.c is only included when SC_HAVE_GETOPT_H is ON.
> 
> maybe we could align autotoools and cmake build in another MR.

I agree, this will be  useful to change and make both systems `#define` the same.

PR #120, opened 2023-04-20
Feature/cmake/build zlib archive

pkestene:
# Add control over building zlib inside libsc

this MR proposes ideas to fix https://github.com/cburstedde/p4est/issues/198 and to issue #96 

we try to fix 2 items:
- slightly change the meaning of cmake option `mpi`; now `mpi=ON` means libsc builds zlib, and `mpi=OFF` means libsc tries to detect zlib from the system. So the user is free to chose if he/she wants to build zlib
-  the second item is more secondary, it adds a cmake option to specifiy a zlib archive on the cmake configure line (to avoid having to download zlib sources). This was motived by the fact a user may want to build p4est with using the website archive (which does not contains zlib sources). The cmake option `zlib`must also exist in p4est to make that possible.

to be discussed further.

### DIFF
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -9,12 +9,26 @@ include(CheckCSourceCompiles)
 if(mpi)
   find_package(MPI COMPONENTS C REQUIRED)
 endif()
+
 if(openmp)
   find_package(OpenMP COMPONENTS C REQUIRED)
 endif()
+
 if(zlib)
+  message(STATUS "Using builtin zlib")
   include(${CMAKE_CURRENT_LIST_DIR}/zlib.cmake)
+  set(SC_HAVE_ZLIB true)
+else()
+  find_package(ZLIB)
+  if(ZLIB_FOUND)
+    message(STATUS "Using system zlib : ${ZLIB_VERSION_STRING}")
+    set(SC_HAVE_ZLIB true)
+  else()
+    message(STATUS "Zlib disabled (not found). Consider using cmake \"-Dzlib=ON\" to turn on builtin zlib.")
+    set(SC_HAVE_ZLIB false)
+  endif()
 endif()
+
 find_package(Threads)
 
 # --- set global compile environment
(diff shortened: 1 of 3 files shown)


cburstedde:
> - slightly change the meaning of cmake option `mpi`; now `mpi=ON` means libsc builds zlib, and `mpi=OFF` means libsc tries to detect zlib from the system. So the user is free to chose if he/she wants to build zlib

Thanks; is it necessary to have one option (MPI) do two things?  I might
prefer having the MPI option behave as closely as possible to autoconf and
to add a ZLIB option just for CMake and building, orthogonal to the MPI.

> to be discussed further.

Appreciate the progress!

pkestene:
> - slightly change the meaning of cmake option `mpi`; now `mpi=ON` means libsc builds zlib, and `mpi=OFF` means libsc tries to detect zlib from the system. So the user is free to chose if he/she wants to build zlib
> Thanks; is it necessary to have one option (MPI) do two things? I might prefer having the MPI option behave as closely as possible to autoconf and to add a ZLIB option just for CMake and building, orthogonal to the MPI.
> to be discussed further.
> Appreciate the progress!

this is a big fat typo, I meant zlib ! Sorry about that. I edited the comment above

cburstedde:
Thanks! I'm just checking with @scivision whether this will be working unchanged with p4est and forestclaw.

StefanBruens:
Now the CMake output is quite awkward:

```
[    2s] -- Found ZLIB: /usr/lib64/libz.so (found version "1.3")  
[    2s] -- Using system zlib : 1.3
...
[    6s] -- The following features have been disabled:
[    6s] 
[    6s]  * MPI, MPI features of libsc
[    6s]  * OpenMP, OpenMP features of libsc
[    6s]  * ZLIB, ZLIB features of libsc
```

The first output is actually correct, sc_io is built with zlib support, and libsc is linked to the system zlib.

PR #127, opened 2023-04-27
update generated pkgconfig (cmake and autotools)

pkestene:
# Update generated pkgconfig (cmake and autotools)

This is a minor change, we just expose 3 variables in pkg-config (autotools and cmake):
- enable_mpi
- build_zlib
- debug_build

these are boolean variables (yes/no) to help a downstream project to quickly get information about how libsc was built.

Another change in the cmake build: we set default value for CMAKE_BUILD_TYPE to RelWithDebInfo for compatibility with autotools defaults.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,13 @@ VERSION ${PROJECT_VERSION})
 
 enable_testing()
 
+include(cmake/init_build_type.cmake)
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+  set(debug_build ON)
+else()
+  set(debug_build OFF)
+endif()
+
 include(cmake/options.cmake)
 message(STATUS "libsc ${PROJECT_VERSION} "
                "install prefix: ${CMAKE_INSTALL_PREFIX}")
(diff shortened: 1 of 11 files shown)


cburstedde:
Thanks; really cool. Should we expand this logic to include ENABLE_MPI_IO?

pkestene:
I need to add json because it is needed in p4est for linking.

Regarding MPIIO, it looks like `SC_ENABLE_MPI`and `SC_ENABLE_MPIIO` are always identical, so I don't know if it really needed, but if you feel it is, we can.

cburstedde:
> I need to add json because it is needed in p4est for linking.

Ok! Same  with zlib presence (do we need to know whether it is built, separately from presence)?

> Regarding MPIIO, it looks like `SC_ENABLE_MPI`and `SC_ENABLE_MPIIO` are always identical, so I don't know if it really needed, but if you feel it is, we can.

In p4est it is possible to configure them differently, so we may create two separate variables. Thanks!

jmark:
> Btw the openmp is on its way out -- we do not expect to expose it far ino the future. Are the t8code folks needing it? @jmark

No, I don't think so. We are not using openmp in t8code.

PR #128, opened 2023-04-30
Update mpiio

pkestene:
# Update MPI-IO mapping

this is minor changes, just update some missing MPI-IO routines.
also sc_MPI_INFO_NULL was defined twice.


src/sc_mpi.h
```diff
@@ -180,7 +180,6 @@ sc_MPI_IO_Errorcode_t;
 #define sc_MPI_STATUSES_IGNORE     MPI_STATUSES_IGNORE
 
 #define sc_MPI_REQUEST_NULL        MPI_REQUEST_NULL
-#define sc_MPI_INFO_NULL           MPI_INFO_NULL
```
cburstedde, line 183: Good catch!

cburstedde:
Thanks; this PR is consistent with libsc currently defining MPI_File_close etc. only when MPIIO is enabled. The new functions do it the same way. In this sense, I see no issue merging, since any code using this file can protect the calls with `#ifdef SC_ENABLE_MPIIO`. What do you think?

pkestene:
@cburstedde 
FYI, I just added a commit because SC_ENABLE_MPIIO was not easily accessible in the cmake build (whereas it was ok in the autotools build). Besides SC_ENABLE_MPIIO was set to 1 when SC_ENABLE_MPI was 1. Not SC_ENABLE_MPIIO is set the same way as in autotools, and can be use in p4est (cmake build).

cburstedde:
Thanks for again aligning CMake some more!

PR #129, opened 2023-06-27
Feature io

cburstedde:
# Update zlib encode/compression convention

## Proposed changes

The compression function for arbitrary array data has been updated to go to 76 code bytes per line and to include a 2-byte general line break. This allows for writing both Unix and MIME type ASCII data.

In passing, remove `sc_io_have_zlib` in favor of `sc_have_zlib`. This is a breaking change requiring a new API version eventually.

### DIFF
--- a/src/sc.c
+++ b/src/sc.c
@@ -1417,8 +1417,6 @@ sc_is_root (void)
 #undef SC_LERRORF
 #endif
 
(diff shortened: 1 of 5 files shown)


tim-griesbach (approved):
This looks good to me!

PR #130, opened 2023-07-13
Make sc_io_open independent of the MPI IO status

tim-griesbach:
# Make `sc_io_open` independent of MPI IO status

This PR is a prerequisite to guarantee that the files produced by the code in https://github.com/cburstedde/p4est/pull/212 are independent of the the MPI IO status.

Proposed changes:
- For the mode `SC_IO_WRITE_CREATE`  for `sc_io_open` we unify the behavior in the case with and without MPI IO. Before these changes an existing file was not truncated to the length zero  with MPI IO. This differed from the case without MPI IO.
- Clarify documentation of `sc_io_open`.
- Fix documentation of `count` parameter of `sc_io` wrapper functions.


cburstedde:
Thanks; very useful!

PR #133, opened 2023-07-18
Add package id query function.

jmark:
In some scenarios it is useful to have a dedicated query function for SC's own package id. For example, not all wrapper
generators or FFIs (foreign function interfaces) support to read global external variables.


cburstedde:
Thanks! I'm wondering, should sc_finalize reset is_initialized to 0.
This will also set the package id back to -1.

The user should be told very clearly that these functions are not thread-safe and refer to global variables.

cburstedde:
Thanks! Just added some doc tweaks.

PR #137, opened 2023-09-21
sc_include.m4: prefix the HAVE_JSON shell variable

cburstedde:
# Use prefix for HAVE_JSON shell variable

Proposed changes: in the autoconf code, move the AC_SUBST macro into the JSON check macro itself.

### DIFF
--- a/config/sc_autotools.pc.in
+++ b/config/sc_autotools.pc.in
@@ -9,8 +9,8 @@ libsc_CFLAGS=@CPPFLAGS@ @CFLAGS@
 
 have_mpi=@HAVE_PKG_MPI@
 have_json=@SC_HAVE_JSON@
-build_zlib=no
 debug_build=@SC_ENABLE_DEBUG@
+build_zlib=no
 
 Name: libsc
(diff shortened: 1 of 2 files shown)


cburstedde:
I'm wondering whether the have_* variables etc. in the pkgconfig file should be prefixed with sc_?

cburstedde:
> This looks good to me!

Thanks for checking!

> > I'm wondering whether the have_* variables etc. in the pkgconfig file should be prefixed with sc_?
> 
> Since we already use `libsc_CC` and `libsc_CFLAGS` in `pkgconf.pc.in` it may be more consistent to also use prefixes for the remaining variables. If we use prefixes for all variables, we may want to decide whether we use `sc_` or `libsc_` throughout the `pkgconf.pc.in` file.

I agree, a common `libsc_` prefix would be nice. @scivision, would there be consequences to prefix have_mpi, have_json, build_zlib and debug_build that we'd need to catch elsewhere?

cburstedde:
Merging for now.
It seems we have little test coverage and user feedback wrt. pkgconfig. Any experts out there?

PR #139, opened 2023-09-30
CMake: Enable valgrind with CTest

### LINKED ISSUE #101: CMake for valgrind with tests
**Description**

With autoconf, we provide the --evable-valgrind option to do two things:

 1. Use valgrind on `make check` if given and valgrind is found.
 2. #define SC_ENABLE_VALGRIND if given and valgrind is found.

Furthermore, p4est will #define P4EST_ENABLE_VALGRIND under the same conditions.
This functionality is, so far, not in the CMake build. How hard would it be to establish it?

elykwilliams:
# Title: Enable Valgrind with CTest

# Resolves #101.

# Overview:
This PR introduces a CMake flag, `TEST_WITH_VALGRIND`, which will run test with valgrind. 

# Details:
- When `TEST_WITH_VALGRIND` is enabled, the configuration attempts to locate Valgrind and defines `SC_ENABLE_VALGRIND=1`.
- Sets the test command as `mpiexec -np 2 valgrind --exit-errorcode=1 sc_test_{}` as with autotools.
- If Valgrind is not enabled, the test command defaults to `mpiexec -np 2 sc_test_{}` (when `mpi=ON` of course).
- A fatal error is emitted by CMake if `TEST_WITH_VALGRIND=ON` but Valgrind is not found.
- Adds workflow to run configure with valgrind enabled and test to mirror the autotools workflow.

### DIFF
--- a/doc/author_williams.txt
+++ b/doc/author_williams.txt
@@ -1,0 +1,1 @@
+I hereby place my contributions to libsc under the FreeBSD license.  Kyle Williams.
--- a/.github/workflows/ci_cmake.yml
+++ b/.github/workflows/ci_cmake.yml
@@ -41,7 +41,7 @@ jobs:
       CC: ${{ matrix.cc }}
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       name: Checkout source code
 
     - name: Install system dependencies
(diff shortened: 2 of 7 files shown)


doc/author_williams.txt
```diff
@@ -0,0 +1 @@
+I hereby place my contributions to libsc into the public domain under FreeBSD license.  Kyle Williams.
```
cburstedde, line 1: Thanks; we've accepted this formulation in the past, but public domain and FreeBSD are not exactly the same. If you were to choose only the latter, that would be perfect.

elykwilliams: Of course, I just copied one of the other ones.

cburstedde:
Thanks! I'd be asking @pkestene to give it a quick look and merge, since I know very little about CMake.

cburstedde:
> [sc.patch](https://github.com/cburstedde/libsc/files/13193238/sc.patch)
> 
> suggested changes to minimize extra variables.

Thanks @scivision! @elykwilliams: please feel free to adjust as you wish.

cburstedde:
Btw. sorry for introducing a conflict by an unrelated merge, but it looks simple to repair.

cburstedde:
I'm hoping this code is current. The CI now does not play along with the valgrind checks, wondering what's up. With `configure --enable-valgrind` I get a clean `make check` on my laptop. What might be going on?

elykwilliams:
Interesting. In previous successful tests, it looks like not all tests were being picked up due to existing checks (only 12 were running as opposed to 14 with autotools). Now, 13 are being picked up with helpers.c being the new test. This "new" is the test that is hanging with with valgrind. 

Currently cmake only adds this test when `SC_SIZEOF_LONG GREATER_EQUAL 8`. I guess this condition when previously running the tests but not now. 

I will look into why this test is hanging, it shouldn't since `make check` works fine. But we should also address why there are different number of tests running between cmake and autotools.

cburstedde:
Thanks for looking into it. sc_test_helpers invokes zlib but can also work without (both configurations run ok with autoconf and valgrind). Wondering whether that's related.

cburstedde:
From the logs, sc_test_helpors links against libz.a under the /home/runner directory, which implies that zlib is downloaded and compiled by the recent CMake logic. Maybe it is built binary incompatible with what valgrind needs? The program seems to abort without further messages.

This is a good example to figure out since we rarely get some coverage for the auto-download of zlib by the CMake logic.

elykwilliams:
It looks like the issue was simply that `test_helpers.c` was taking close to 30 seconds to run, and CTest had a default test timeout set at 30 seconds. I modified the `tests/CMakeLists.txt` to set the timeout at 60 seconds per test. The tests are now passing fine, I've gone ahead and squashed all the subsequent commits into a single commit.

cburstedde:
Ok this sounds good! Thanks again, merging.

PR #140, opened 2023-10-04
CMake: corrections for Zlib and Rpath

scivision:
* Simplify/correct specification of Zlib local archive vs. remote download if Zlib build required
* RPath: use CMake defaults as they are better in general for complex situations like HPC module loading
* bump CMake max version to avoid deprecation warning with recent CMake releases
* correct default CMAKE_BUILD_TYPE so it actually takes effect and respects env var CMAKE_BUILD_TYPE if set


cburstedde:
Just checking real quick: this uses the system zlib by default as desired by Linux maintainers?

PR #141, opened 2023-10-04
CMake: correct Jansson

scivision:
NOTE: requires PR #140 because so many  changes were made in #140

To greatly simplify use of Jansson in libsc CMake, require that Jansson was installed by CMake. This also greatly simplifies use of JSON from p4est and other upstream libsc-consuming CMake packages.

If desired, I could make Jansson be build via ExternalProject like Zlib (I didn't do that yet, in case you didn't want it).


cburstedde:
Thanks again!

PR #143, opened 2023-11-01
CMake: remove unused packaging script

scivision:
FindSC.cmake was never used by the CMake package, it was always superfluous, so remove it.


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14...3.28)
+cmake_minimum_required(VERSION 3.15...3.28)
```
elykwilliams, line 1: Why does this commit bump the minimum required version?

elykwilliams:
Why bump the cmake min version number? Am I missing the change that requires 3.15?

cburstedde:
Thanks! Especially the cmakedefine for zlib will be a good thing. Would it be possible to make this PR not from the ECLAIRWaves organization, or to rebase it against @cburstedde's github repository? It has a develop branch that does not match the branch here.

cburstedde:
Ok, using this repository now. Thanks!

PR #145, opened 2023-11-18
Simplify MPI/IO wrappers

cburstedde:
# Simplify MPI/IO wrappers

## Proposed changes

The sc_io files wrap several MPI file functions to account for configuration with and without MPI, and with MPI but without MPI I/O found/selected. These configurations are distinguished at compile time, and in this PR we simplify the code to leave as little content inside of ifdefs as possible. Along the way, we have some minor edits for assertions and variable names.

There is one fix to free the file structure on file open error return. Wondering whether this is sufficiently documented.

What's left is to recheck the standards (see TO DO in the comments). For the functions that do not move the MPI file pointer, the non-MPI replacements should take care to replicate this behavior with ftell/fseek as closely as possible.

### DIFF
--- a/doc/release_notes.txt
+++ b/doc/release_notes.txt
@@ -31,6 +31,7 @@ There have been breaking changes, very strictly speaking.
 ### Functionality
 
  - Further align I/O wrappers with and without MPI I/O.
+ - Simplify the replacements for MPI_{read,write}_at*.
  - Update zlib and base64 encode/compression routines.
 
 ## 2.8.5
--- a/src/sc_io.c
+++ b/src/sc_io.c
@@ -1367,7 +1367,7 @@ sc_io_parse_access_mode (sc_io_open_mode_t amode, const char **mode)
 
 #else
 
-typedef int sc_io_access_mode_t;
+typedef int         sc_io_access_mode_t;
 
 static void
 sc_io_parse_access_mode (sc_io_open_mode_t amode, int *mode)
@@ -1401,10 +1401,10 @@ sc_io_open (sc_MPI_Comm mpicomm, const char *filename,
 {
   sc_io_access_mode_t mode;
   int                 mpiret, errcode, retval;
-#ifdef SC_ENABLE_MPIIO
 
   sc_io_parse_access_mode (amode, &mode);
 
+#ifdef SC_ENABLE_MPIIO
   mpiret = MPI_File_open (mpicomm, filename, mode, mpiinfo, mpifile);
   retval = sc_io_error_class (mpiret, &errcode);
   SC_CHECK_MPI (retval);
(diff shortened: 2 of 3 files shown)


cburstedde:
We may need to double check that each error return is synchronized with MPI but without MPI I/O. I'm suspecting my sc_io_close needs fixing, may there be other occurrences?

Please feel free to add to the top history to address this and the remaining TO DO items in the code.

tim-griesbach:
> There is one fix to free the file structure on file open error return. Wondering whether this is sufficiently documented.

I added a clarification about this in the documentation in `sc_io_open`.

> What's left is to recheck the standards (see TO DO in the comments). For the functions that do not move the MPI file pointer, the non-MPI replacements should take care to replicate this behavior with ftell/fseek as closely as possible.

The TO DO comments are addressed in the commits that I pushed on the top of this PR. I left a comment above on the TO DO comment on the fflush in `sc_write_at`.


> We may need to double check that each error return is synchronized with MPI but without MPI I/O. I'm suspecting my sc_io_close needs fixing, may there be other occurrences?

I added the synchronization of the error return values for the functions `sc_io_close`, `sc_io_write_at` and `sc_io_read_at`.

cburstedde:
> > We may need to double check that each error return is synchronized with MPI but without MPI I/O. I'm suspecting my sc_io_close needs fixing, may there be other occurrences?
> 
> I added the synchronization of the error return values for the functions `sc_io_close`, `sc_io_write_at` and `sc_io_read_at`.

Thanks for the quick response!  I'm only wondering whether this is the right
thing to do for the latter two (non-collective) functions.

tim-griesbach:
Following our discussion of the code the new commits
- add the file error checking in `sc_io_{write,read}_at` in the non-MPI-IO case,
- return `sc_MPI_ERR_ARG` as error code in case of illegal usage of `sc_io_{write,read}_at`,
- introduce  `sc_io_{write,read}_at_legal` to check if the restrictions of writing/reading with `zcount > 0` only on rank 0 hold for the respective function and
- document the changes and add the recommendation to overcome restrictions of `sc_io_{write,read}_at` using `sc_io_{write,read}_at_all`.

src/sc_io.h
```diff
@@ -562,6 +571,17 @@ void                sc_io_read (sc_MPI_File mpifile, void *ptr,
                                 size_t zcount, sc_MPI_Datatype t,
                                 const char *errmsg);
 
+/** Check for restricted usage of \ref sc_io_read_at.
+ *
+ * \return              0 if the restriction described in the note of \ref
+ *                      sc_io_read_at applies. This is equivalent to MPI I/O
```
cburstedde, line 577: Would you be self-consistent here and briefly but completely state the requirement, so the reader does not have to go looking for the original definition?

cburstedde:
Merging into prev3-develop since that is the latest branch with the p4est develop CI going through.

Issue #134, opened 2023-07-18
project version is 0.0.0 when compiling libsc from git tarball


topazus:
**Description**

<details>
<summary>build log</summary>

```bash
ruby in 🌐 fedora in rpmbuild/SOURCES/libsc-83bd019fc9f029a9c15a3ae823e6d7d20e6e0340 via △ v3.27.0-rc4 
❯ cmake . -B build -GNinja -DCMAKE_BUILD_TYPE=Release -Dzlib=OFF -DCMAKE_SKIP_RPATH=ON -DBUILD_SHARED_LIBS=ON && cmake --build build -j6
-- The C compiler identification is GNU 13.1.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/lib64/ccache/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- libsc 0.0.0 install prefix: /home/ruby/rpmbuild/SOURCES/libsc-83bd019fc9f029a9c15a3ae823e6d7d20e6e0340/build/local
-- Found ZLIB: /usr/lib64/libz.so (found version "1.2.13")  
-- Using system zlib : 1.2.13
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Could NOT find jansson (missing: jansson_DIR)
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.9.5") 
-- jansson library found via pkg-config
-- Looking for sqrt
-- Looking for sqrt - not found
-- Looking for sqrt
-- Looking for sqrt - found
-- Looking for realloc
-- Looking for realloc - found
-- Looking for aligned_alloc
-- Looking for aligned_alloc - found
-- Looking for backtrace
-- Looking for backtrace - found
-- Looking for backtrace_symbols
-- Looking for backtrace_symbols - found
-- Looking for execinfo.h
-- Looking for execinfo.h - found
-- Looking for fsync
-- Looking for fsync - found
-- Looking for inttypes.h
-- Looking for inttypes.h - found
-- Looking for memory.h
-- Looking for memory.h - found
-- Looking for posix_memalign
-- Looking for posix_memalign - found
-- Looking for basename
-- Looking for basename - found
-- Looking for qsort_r
-- Looking for qsort_r - found
[...]

PR #147, opened 2023-12-18
Fix the SONAME in CMake to conform to that in libtool

topazus:
related: https://github.com/cburstedde/libsc/issues/134#issuecomment-1729115323

The results of symbolic links with CMake

```bash
❯ ll ./build/libsc.so*
lrwxrwxrwx. 1 aurora aurora     10 Dec 18 23:10 ./build/libsc.so -> libsc.so.2
lrwxrwxrwx. 1 aurora aurora     14 Dec 18 23:10 ./build/libsc.so.2 -> libsc.so.2.0.0
-rwxr-xr-x. 1 aurora aurora 712032 Dec 18 23:10 ./build/libsc.so.2.0.0
```

build log:
<details>

```bash
❯ rm -rf build && cmake . -B build -DBUILD_SHARED_LIBS=ON -Dzlib=OFF
-- Set build type to 'RelWithDebInfo' as none was specified.
-- The C compiler identification is GNU 13.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/lib64/ccache/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- libsc 2.0.0 install prefix: /home/aurora/fedora-src/libsc/build/local
-- Found ZLIB: /usr/lib64/libz.so (found version "1.2.13")  
-- Using system zlib : 1.2.13
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Could NOT find jansson (missing: jansson_DIR)
-- Looking for sqrt
-- Looking for sqrt - not found
-- Looking for sqrt
-- Looking for sqrt - found
-- Looking for realloc
-- Looking for realloc - found
-- Looking for aligned_alloc
-- Looking for aligned_alloc - not found
-- Looking for _aligned_malloc
-- Looking for _aligned_malloc - not found
-- Looking for backtrace
-- Looking for backtrace - found
-- Looking for backtrace_symbols
-- Looking for backtrace_symbols - found
-- Looking for execinfo.h
-- Looking for execinfo.h - found
-- Looking for fsync
-- Looking for fsync - found
-- Looking for inttypes.h
-- Looking for inttypes.h - found
-- Looking for memory.h
-- Looking for memory.h - found
-- Looking for posix_memalign
-- Looking for posix_memalign - found
-- Looking for basename
-- Looking for basename - found
-- Looking for qsort_r
-- Looking for qsort_r - found
-- Checking prototype qsort_r for SC_HAVE_GNU_QSORT_R
-- Checking prototype qsort_r for SC_HAVE_GNU_QSORT_R - True
-- Checking prototype qsort_r for SC_HAVE_BSD_QSORT_R
-- Checking prototype qsort_r for SC_HAVE_BSD_QSORT_R - False
-- Looking for fabs
-- Looking for fabs - found
-- Looking for signal.h
-- Looking for signal.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for stdlib.h
-- Looking for stdlib.h - found
-- Looking for string.h
-- Looking for string.h - found
-- Looking for random
-- Looking for random - found
-- Looking for srandom
-- Looking for srandom - found
-- Looking for strtoll
-- Looking for strtoll - found
-- Looking for strtok_r
-- Looking for strtok_r - found
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for sys/time.h
-- Looking for sys/time.h - found
-- Looking for time.h
-- Looking for time.h - found
-- Looking for libgen.h
-- Looking for libgen.h - found
-- Looking for unistd.h
-- Looking for unistd.h - found
-- Looking for getopt.h
-- Looking for getopt.h - found
-- Looking for sys/ioctl.h
-- Looking for sys/ioctl.h - found
-- Looking for sys/select.h
-- Looking for sys/select.h - found
-- Looking for sys/stat.h
-- Looking for sys/stat.h - found
-- Looking for fcntl.h
-- Looking for fcntl.h - found
-- Looking for linux/videodev2.h
-- Looking for linux/videodev2.h - found
-- Looking for linux/version.h
-- Looking for linux/version.h - found
-- Performing Test SC_ENABLE_V4L2
-- Performing Test SC_ENABLE_V4L2 - Success
-- Check size of int
-- Check size of int - done
-- Check size of unsigned int
-- Check size of unsigned int - done
-- Check size of long
-- Check size of long - done
-- Check size of long long
-- Check size of long long - done
-- Check size of unsigned long
-- Check size of unsigned long - done
-- Check size of unsigned long long
-- Check size of unsigned long long - done
-- Looking for SC_ENABLE_MPI
-- Looking for SC_ENABLE_MPI - not found
-- Looking for SC_ENABLE_MPIIO
-- Looking for SC_ENABLE_MPIIO - not found
-- The following features have been enabled:

 * ZLIB, ZLIB features of libsc
 * shared, shared libsc library

-- The following features have been disabled:

 * MPI, MPI features of libsc
 * OpenMP, OpenMP features of libsc
 * json, JSON features of libsc

-- Configuring done (3.3s)
-- Generating done (0.0s)
-- Build files have been written to: /home/aurora/fedora-src/libsc/build
```

<details>


cburstedde:
Thanks! As for nomenclature, with autoconf we set the libtool version to, say, 2.0.0 and bump it according to the libtool documentation, while the release version is something like v2.8.6 and is bumped by increasing one of the digits.

Would it make sense to differentiate between the two in CMake by the name of the variables used and/or comments?

topazus:
Some project set the same project version in CMake as the github release version, like yaml-cpp

https://github.com/jbeder/yaml-cpp/blob/c7639e81d5f00a5b47a2c9bd668f10d74b949071/CMakeLists.txt#L145-L151

Some project differentiate them, like simdjson

https://github.com/simdjson/simdjson/blob/ebd09cb2a3ab421f04b404338c01b593ade5a08c/CMakeLists.txt#L22-L24

cburstedde:
I think I prefer the differentiated case. LIBSC_VERSION = LIBSC_PACKAGE_VERSION = version based on vx.y.z tags in the repository (ideally generated by CMake using the same script as autoconf: build_aux/git-version-gen).

Separately, LIBSC_SOVERSION may be set in the CMake files to the -version-info from src/Makefile.am.

 If on top of this, the CI would run through, that would be perfect. Thanks for your input!

topazus:
I fix this, but I may not modified as you said because I do not fully understand your point.

cburstedde:
Thanks; I might be adding to this PR in a bit.

Would you mind adding a file doc/author_<lastname>.txt releasing your code under the FreeBSD license?

cburstedde:
I've used the file mechanism suggested in #134. Ideally, both autoconf and cmake should be producing the same release version (aka tarball version) as well as the same library version (aka soname). What do you think about this, will it work?

cburstedde:
With the same build options as above, I'm now getting
Dec 20 14:50 libsc.so -> libsc.so.2.0.0
Dec 20 14:50 libsc.so.2.0.0 -> libsc.so.2.8.5.999
Dec 20 14:50 libsc.so.2.8.5.999
This seems not ideal, mixing release and so versions in the library names.
How can we fix this to use only the soname in the .so files/links?

cburstedde:
Ok, I'm using *only* SOVERSION and this seems to clean everything up. Thanks for your input!
Anything else to be done within this PR or #134?

topazus:
The behaviors about the versioning and symlinks of the shared library files seems to be different between CMake build and autotools build.

Ref: https://github.com/zeromq/libzmq/issues/2887

> In most projects that got ported from an environment that uses GNU libtool (for example autotools) to for example cmake or meson, and in the rare cases that they did anything at all in a qmake based project, I saw people converting the current, revision and age parameters that they passed to the -version-info option of libtool to a string concatenated together using (current - age), age, revision as VERSION, and (current - age) as SOVERSION.
>
> I wanted to use the exact same rules for versioning for all these examples, including autotools and GNU libtool. When you don't have to (or want to) care about libtool's set of (for some people, needlessly complicated) -version-info rules, then it should be fine using just SOVERSION and VERSION using these rules:
>
> SOVERSION = Major version
Major version: increase it if you break ABI compatibility
Minor version: increase it if you add ABI compatible features
Patch version: increase it for bug fix releases.

For libzmq, the version in libtool needs some calculation to transform into SOVERSION and VERSION in CMake.
- libtool: 7:6:2
- CMake: VERSION is (7-2):2:6 ,so "5.2.6". SOVERSION is 7-2 , so "5".

Ref: https://github.com/zeromq/libzmq/blob/959a133520dfc80d29e83aa7ef762e1d0327f63b/configure.ac#L59
https://github.com/zeromq/libzmq/blob/959a133520dfc80d29e83aa7ef762e1d0327f63b/CMakeLists.txt#L1396
https://github.com/pvanhoof/dir-examples?tab=readme-ov-file#what-is-this-soversion-about

cburstedde:
Thanks for your comments! I'm going for the simplest way of maintaining the soversion by hand.

PR #148, opened 2023-12-21
find Zlib and test: ensure adler32_combine available

scivision:
It is desired to find Zlib as most personal computers (laptops, workstations) have a new-enough Zlib, and Linux distributions in general like Debian likewise have new-enough Zlib.

To avoid the issue with `find_package(ZLIB)` not checking for `adler32_combine()` and declaring imported target `ZLIB::ZLIB`, we instead search for the same Zlib binary file names as find_package(ZLIB) but run an adler32_combine quick check before declaring ZLIB::ZLIB.


cburstedde:
Thanks again! I'm closing this one since it's coming from ECLAIRWaveS. Happy to go back to the previous PR workflow.

Issue #121, opened 2023-04-21
Slight change in the cmake install procedure


pkestene:
**Description**

Currenly when building and installing p4est/libsc with cmake there is a difference between Debian-like and RedHat-like systems:

- on a Debian-like system libp4est.so and libsc.so are installed in `${CMAKE_INSTALL_PREFIX}/lib/` => ok
- on  a RedHat-like system libp4est.so is installed in  `${CMAKE_INSTALL_PREFIX}/lib/`  and libsc.so gets installed in  `${CMAKE_INSTALL_PREFIX}/lib64/` ; this is a bit annoying, and introduced a discrepancy with the autotools build where all lib are always installed in   `${CMAKE_INSTALL_PREFIX}/lib/` 

The reason why is explained cmake` install` command doc: https://cmake.org/cmake/help/latest/command/install.html
In libsc, when macro `install(TARGETS sc EXPORT ${PROJECT_NAME}-targets)` is called without specifying ARCHIVE/LIBRARY/RUNTIME), the target gets installed in the built-in default path (lib) or in CMAKE_INSTALL_LIBDIR (if defined by macro GNUInstalldirs).

Actually what happends here, is that in libsc GNUInstallDirs macro is loaded, but not in p4est.
So libsc is installed in CMAKE_INSTALL_LIBDIR (which is lib64 in RedHat-like OS, and lib in Debian-like OS).
While p4est.so is always installed in lib, indepenent of the OS.

**Proposed solution**

As there is no equivalent to GNUInstallDirs in autotools (to my knowledge), and as autotools always installs libraries in "lib" subfolder, I suggest the following change in libsc top-level CMakeLists.txt

```cmake
install(TARGETS sc EXPORT ${PROJECT_NAME}-targets)
```
into
```cmake
install(TARGETS sc
  EXPORT ${PROJECT_NAME}-targets
  ARCHIVE DESTINATION lib
  LIBRARY DESTINATION lib
  RUNTIME DESTINATION bin)
```
so that when libsc is built inside p4est with cmake, whatever the OS, all libs (libp4est.so and libsc.so) are both installed in lib folder.

cburstedde:
Thanks! This is a change for the better.

With the autotools, the install paths can be differentiated using

    --bindir=
    --libexecdir=
    --libdir=

etc. with sensible defaults as you described. So the flexibility is there, generally.

cburstedde:
This seems complete?

pkestene:
[...]

PR #149, opened 2023-12-22
Zlib: bugfix lib64 for RHEL/SUSE/Cray, Zlib 2.1.5, test for adler32_combine

scivision:
* RHEL/SUSE/Cray platforms and others using lib64 were broken when building zlib. This fixes that. Relevant to discussion in #121 (but independent of that)
* use latest Zlib 2.1.5
*  only use found Zlib if adler32_combine is present
* given above, make default to build Zlib OFF
* lint: use SC_IS_TOP_LEVEL to be more explicit in intent
* partly revert #122 relevant to CMake packaging to use the default GNUInstallDirs that other CMake packages expect, particularly for RHEL/SUSE/Cray systems


cburstedde:
Thanks! Pulling it in now.

PR #150, opened 2023-12-24
Intel oneAPI Windows: add support

scivision:
Intel oneAPI works on Windows with this PR, it did not previously. The CMake logic is also corrected to only use the custom gettimeofday when needed as MinGW doesn't need it.

SC_HAVE_WINSOCK2_H was no longer needed and hence removed

Microsoft VIsual Studio itself still doesn't work due to the `__attribute__` properties that are not supported by Visual Studio.

This was tested with oneAPI Windows with and without MPI.


cburstedde:
Thanks; looks really nice!

PR #152, opened 2023-12-24
make logic more programmatic

scivision:
rather than blanket ifdef _MSC_VER,
 the programmatic define for libgen.h is adequate.

For MSVC, rather than not sleep, use the Sleep equivalent to sleep

This is not related to the other PRs currently open.


cburstedde:
Cool, I like the simplification.

PR #154, opened 2024-01-03
cmake:SC_ENABLE_MPI cache var like SC_HAVE_ZLIB for global scope

scivision:
A global scope (e.g. cache) variable is necessary for p4est, forestclaw, etc. to properly propagate libsc configuration for MPI

use SC_ENABLE_MPI and SC_HAVE_ZLIB rather than option() variables for consistent logic.

### DIFF
--- a/cmake/pkgconf.cmake
+++ b/cmake/pkgconf.cmake
@@ -1,8 +1,8 @@
 # --- generate pkg-config .pc
 set(pc_req_private "")
-if(mpi)
+if(SC_ENABLE_MPI)
   string(APPEND pc_req_private " ompi ompi-c zlib")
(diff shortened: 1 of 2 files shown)


cmake/pkgconf.cmake
```diff
@@ -1,8 +1,8 @@
 # --- generate pkg-config .pc
 set(pc_req_private "")
-if(mpi)
+if(SC_ENABLE_MPI)
   string(APPEND pc_req_private " ompi ompi-c zlib")
```
cburstedde, line 4: Thanks; unrelated to this PR: are we depending on a particular (OpenMPI) MPI installation?

PR #157, opened 2024-01-16
Feature sink source

cburstedde:
# Improvements to the sc_io source/sink functionality 

There are quite a few comment and documentation updates and two functional ones:

 1. We remember the end-of-file status on reading from a buffer or file and exit early next time around.
 2. When we read from an array buffer, we read at most its current elements and not, as previously, to the end of its internal memory allocation, which may be larger and contain undefined data.

In particular the latter update may have implications to existing code!


tim-griesbach (commented):
This looks good to me!

PR #159, opened 2024-01-19
CMake: add Jansson like Zlib

scivision:
Jansson would then be visible in p4est etc. without changes to those projects (once p4est updates its libsc Git submodule).


cburstedde:
Thanks; much appreciated! There's a (transient?) CI fail though, wondering why that always seems to happen on MacOS.

Referring to @tim-griesbach for further comments.

tim-griesbach:
> Referring to @tim-griesbach for further comments.

Thanks to @scivision  for the quick PR creation! I tried out the branch and it works for jansson analogously to zlib, i.e. if jansson is not found, it is deactivated but the user can activate an auto build of jansson. In the auto build case, I could successfully read a jansson file.

PR #160, opened 2024-01-19
Feature hash array

cburstedde:
# Complete hash array internal state update

## Proposed changes

We had begun to hide hash array internal state in 8e3ff6fced3353cd0e9e7d6f767ad2010235069a.
This breaks p4est since it accesses some internal data. p4est will be fixed after merging this PR.

In this PR we add sc_hash_array_foreach to avoid accessing obscure internal variables.
Me add some minor tweaks to sc_hash and sc_hash_array.

### DIFF
--- a/src/sc_containers.c
+++ b/src/sc_containers.c
@@ -1523,7 +1523,7 @@ struct sc_hash_array_data
   sc_array_t         *pa;
   sc_hash_function_t  hash_fn;
   sc_equal_function_t equal_fn;
-  void               *user_data;
+  sc_hash_foreach_t   foreach_fn;
   void               *current_item;
 };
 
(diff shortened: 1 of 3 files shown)


src/sc_containers.c
```diff
@@ -1631,6 +1637,11 @@ sc_hash_array_lookup (sc_hash_array_t * hash_array, void *v, size_t *position)
   int                 found;
   void              **found_void;
 
+  /* verify general invariant */
+  SC_ASSERT (hash_array != NULL);
+  SC_ASSERT (hash_array->a.elem_count == hash_array->h->elem_count);
+  SC_ASSERT (hash_array->internal_data->foreach_fn == NULL);
```
hannesbrandt, line 1643: Would it make sense to check for `hash_array->internal_data->v == NULL` here too, since it is set to NULL after every usage, similar to `foreach_fn`?

hannesbrandt:
Looks good to me!

src/sc_containers.c
cburstedde: Thanks for pushing the addition!

PR #162, opened 2024-01-24
CMake: redo check when "mpi" option changes

scivision:
This allows the user to toggle "mpi" on and off without completely clearing the build directory.


cburstedde:
Thanks let's try this with p4est.

PR #164, opened 2024-01-24
CMake: remove OpenMP

scivision:
At your convenience (nothing is depending on this) it looks like OpenMP was never really used in CMake libsc. If so, just removing it to not have unused options in the script.


cburstedde:
I think this is worth it, let's simplify for lack of support of OpenMP.

PR #168, opened 2024-02-15
Feature MPI_ERR_COUNT: was it missing before?

cburstedde:
# Add sc_MPI_ERR_COUNT

Proposed changes: The MPI_ERR_COUNT value was not there. Adding to our MPI compatibility layer.

Please comment if this is not the right thing, otherwise merge right away @tim-griesbach .


tim-griesbach:
It is a good point that MPI_ERR_COUNT is partially missing in our MPI compatibility layer. It was already present in the enum `sc_MPI_IO_Errorcode` in `sc_mpi.h` for the case without MPI I/O.

However, I think that this way of dealing with MPI_ERR_COUNT is not entirely correct. Since the MPI standard (https://www.mpi-forum.org/docs/mpi-2.2/mpi22-report/node192.htm#Node192) says
```
MPI_ERR_COUNT	Invalid count argument 
```
To my understanding, this means that `MPI_ERR_COUNT` can also occur without MPI I/O for all MPI functions with a count argument. Therefore, I would propose the following changes:
- Remove `sc_MPI_ERR_COUNT` from the enum `sc_MPI_IO_Errorcode`,
- add `#define sc_MPI_ERR_COUNT MPI_ERR_COUNT` in the block of MPI error constants like for MPI case,
- add `#define sc_MPI_ERR_COUNT SC3_MPI_ERR_COUNT` for the case without MPI and
- add `MPI_ERR_COUNT` to `sc3_mpi_types.h` as an error code that is also relevant for the MPI case.

These changes would ensure that `MPI_ERR_COUNT` is correctly handled in our MPI compatibility layer independent of MPI I/O being available and activated.

If you agree with these suggestions, I could push these changes to the top of this PR.

cburstedde:
> It is a good point that MPI_ERR_COUNT is partially missing in our MPI compatibility layer. It was already present in the enum `sc_MPI_IO_Errorcode` in `sc_mpi.h` for the case without MPI I/O.
> 
> However, I think that this way of dealing with MPI_ERR_COUNT is not entirely correct. Since the MPI standard (https://www.mpi-forum.org/docs/mpi-2.2/mpi22-report/node192.htm#Node192) says
> ```
> MPI_ERR_COUNT	Invalid count argument 
> ```
> To my understanding, this means that `MPI_ERR_COUNT` can also occur without MPI I/O for all MPI functions with a count argument. Therefore, I would propose the following changes:
> - Remove `sc_MPI_ERR_COUNT` from the enum `sc_MPI_IO_Errorcode`,
> - add `#define sc_MPI_ERR_COUNT MPI_ERR_COUNT` in the block of MPI error constants like for MPI case,
> - add `#define sc_MPI_ERR_COUNT SC3_MPI_ERR_COUNT` for the case without MPI and
> - add `MPI_ERR_COUNT` to `sc3_mpi_types.h` as an error code that is also relevant for the MPI case.
> 
> These changes would ensure that `MPI_ERR_COUNT` is correctly handled in our MPI compatibility layer independent of MPI I/O being available and activated.
> 
> If you agree with these suggestions, I could push these changes to the top of this PR.

Thanks for the report!  ERR_COUNT seems to appear in MPI 1.3, so I would
agree to move it outside into the general section.

tim-griesbach:
> Thanks for the report! ERR_COUNT seems to appear in MPI 1.3, so I would agree to move it outside into the general section.

I pushed a commit that moves `ERR_COUNT` to the general MPI error code section.

Issue #170, opened 2024-02-23
development branch does not compile with MPITrampoline


jmark:
The development branch cannot be compiled with the MPITrampoline ABI wrapper:
https://github.com/eschnett/MPItrampoline

I specifically need this to provide a broad support of platforms for our t8code Julia bindings.

I describe the problem with a specific example:

Within `mpi.h` provided by MPITrampoline `MPI_ERR_LASTCODE` is not a real compile-time constant. This is true
for all other MPI constants as well.

```
// mpi.h
extern int MPITRAMPOLINE_CONST MPI_ERR_LASTCODE
```

In `libsc` `MPI_ERR_LASTCODE` is used in enums:

```
// sc_mpi.h
#define sc_MPI_ERR_LASTCODE               MPI_ERR_LASTCODE`
```
and
```
// sc_sda.h
typedef enum sc_scda_ret
{
  SC_SCDA_FERR_SUCCESS = 0, /**< successful function call */
  SC_SCDA_FERR_FORMAT = sc_MPI_ERR_LASTCODE,  /**< File not conforming to the
                                                   \b scda format. */
[...]
};
```

This gives the following error:
![grafik](https://github.com/cburstedde/libsc/assets/10619309/cb01ae42-7b93-4741-8ed3-6e2b2049e1dd)

@lukasdreyer

cburstedde:
Thanks;, let's discuss with @tim-griesbach how we may address this.

jmark:
There is also a discussion going at the MPTrampoline side: https://github.com/eschnett/MPItrampoline/issues/37

cburstedde:
We're working on remiving the use of MPI_ERR_LASTCODE as a constant.

We'll alse be deprecating the configuration with MPI but without MPI I/O, leaving either non-MPI operation or MPI operation with full MPI_File support. Any opinions on this step?

cburstedde:
I've just merged work by @tim-griesbach not to rely on MPI_ERR_LASTCODE anymore.

We have deprecated the combo --enable-mpi --disable-mpiio. Does all else work for y'all?

cburstedde:
> I've just merged work by @tim-griesbach not to rely on MPI_ERR_LASTCODE anymore.
> 
> We have deprecated the combo --enable-mpi --disable-mpiio. Does all else work for y'all?

We have merged a fix for these cases. About to release 2.8.6! Any comments?

PR #171, opened 2024-03-06
Deprecate MPI without MPI I/O

tim-griesbach:
# Deprecate MPI without MPI I/O

Proposed changes:
- Deprecate the configuration with MPI and without MPI I/O.
- Avoid using MPI link-time constants as compile-time constants in non-deprecated configuration setups.
- Exclude the case of MPI without MPI I/O from the documentation in `sc_io.h`.
- Remove `sc_io_{write,read}_at_legal` since they are only related to the deprecated configuration setup.
- Remove `sc_io_{write,read}_all` since it is unused and not conforming to the corresponding `MPI_File` functions in any configuration setup.
- Update the release notes for the changes in this PR, change existing release notes concerning MPI without MPI I/O and adding a note for the PR https://github.com/cburstedde/libsc/pull/130.

This draft PR would also profit from the changes in @cburstedde's branch https://github.com/cburstedde/libsc/tree/feature-disable-mpiio introducing warnings if the deprecated configuration setup is used.

Moreover, this draft PR relates to the issue https://github.com/cburstedde/libsc/issues/170 by not using `MPI_ERR_LASTCODE` in non-deprecated configuration setups.

### DIFF
--- a/src/sc_scda.h
+++ b/src/sc_scda.h
@@ -304,7 +304,6 @@ sc_scda_fopen_options_t; /**< type for \ref sc_scda_fopen_options */
  * [preprint](https://doi.org/10.48550/arXiv.2307.06789).
  *
  * This function returns NULL on MPI I/O errors.
- * Without MPI the function may abort on file system dependent errors.
  *
  * \param [in]     mpicomm   The MPI communicator that is used to open the
  *                           parallel file.
@@ -348,8 +347,6 @@ sc_scda_fcontext_t *sc_scda_fopen_write (sc_MPI_Comm mpicomm,
  * All parameters except of \b data are collective.
  *
  * This function returns NULL on MPI I/O errors.
- * Without MPI the function may abort on file system dependent
- * errors.
  *
  * \param [in,out]  fc          File context previously opened by \ref
  *                              sc_scda_fopen_write.
@@ -394,8 +391,6 @@ sc_scda_fcontext_t *sc_scda_fwrite_inline (sc_scda_fcontext_t * fc,
  * All parameters except of \b block_data are collective.
  *
  * This function returns NULL on MPI I/O errors.
- * Without MPI the function may abort on file system dependent
- * errors.
  *
  * \param [in,out]  fc          File context previously opened by \ref
  *                              sc_scda_fopen_write.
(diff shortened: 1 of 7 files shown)


cburstedde:
Thanks. Will it be reasonable to integrate #172 first?

tim-griesbach:
> Thanks. Will it be reasonable to integrate #172 first?

Yes, I think it makes sense to integrate the PR https://github.com/cburstedde/libsc/pull/172 first.

src/sc_scda.h
```diff
@@ -304,7 +304,7 @@ sc_scda_fopen_options_t; /**< type for \ref sc_scda_fopen_options */
  * [preprint](https://doi.org/10.48550/arXiv.2307.06789).
  *
  * This function returns NULL on MPI I/O errors.
- * Without MPI I/O the function may abort on file system dependent errors.
+ * Without MPI the function may abort on file system dependent errors.
```
cburstedde, line 307: A question not for this PR, just to keep in mind: will we have non-abort error checks without MPI?

cburstedde:
Thanks! I'm just wondering whether the scda feature should do full error checking both with and without MPI. Is it necessary to document that we may abort? This limits community use.

tim-griesbach:
> Thanks! I'm just wondering whether the scda feature should do full error checking both with and without MPI. Is it necessary to document that we may abort? This limits community use.

This is a very good point! I rechecked the code and without the deprecated case of MPI without MPI I/O we only have one file functions related abort. It is just a simple consistency check in `sc_io_close`. 
```
    errno = 0;
    retval = fclose ((*mpifile)->file);
    mpiret = sc_io_error_class (errno, &eclass);
    SC_CHECK_MPI (mpiret);
    SC_CHECK_ABORT (!retval == (eclass == sc_MPI_SUCCESS),
                    "fclose return value inconsistent");
```

Therefore, we may want to remove the comments on the file system dependent aborts since there will be no aborts beyond simple consistency checks.

src/sc_scda.h
```diff
@@ -394,8 +391,6 @@ sc_scda_fcontext_t *sc_scda_fwrite_inline (sc_scda_fcontext_t * fc,
  * All parameters except of \b block_data are collective.
  *
  * This function returns NULL on MPI I/O errors.
```
cburstedde, line 393: Very strictly speaking, should we just write "on I/O errors" to account for the non-MPI case?

tim-griesbach: Yes, this is more accurate.

cburstedde:
Thanks! This migth be all for this PR.

When continuing working with SCDA, we may comment on other contents that may occur, mostly errors in the file contents encountered on reading.

cburstedde:
Are you generally happy with the state of our doxygen output?

tim-griesbach:
> Are you generally happy with the state of our doxygen output?

The doxygen output looks good to me. I adjusted the I/O page since it was still referring to the deprecated case with MPI but without MPI I/O.

cburstedde:
Cool let's see how this flies with the community once p4est updates its submodule.

PR #172, opened 2024-03-07
Feature disable mpiio

cburstedde:
# Deprecate `--disable-mpiio` configure option

Use of MPI configuration without MPI I/O enabled now results in a warning and DEPRECATED message.
It has become increasingly expensive to maintain this case with reasonable standards conformance.
We are suggesting this PR as a first step to more drastically disable this corner case.


tim-griesbach (commented):
This looks good to me!

PR #174, opened 2024-03-11
CI: macOS: CMake: use Apple Silicon hardware

scivision:
Most Apple users now have Apple Silicon ARM64 CPU. Since Apple Silicon introduction, there have been several Apple Silicon-specific issues that needed tweaking. In Jan. 2024, GitHub Actions made Apple Silicon CI free to use so this PR uses that to help detect future issues.

Previously, GitHub Actions used macOS 13 on Intel CPU, which missed numerous issues our developers faced on their Apple Silicon laptops.


cburstedde:
Thanks!

PR #175, opened 2024-03-14
Simplify MPI data types to old standard

cburstedde:
We had a couple data types that came in after MPI 1.3. Some were not consistently represented in all their aspects. We removed sc_MPI_SIGNED_CHAR and sc_MPI_DOUBLE_INT. We redefined INT8 to BYTE and UNSIGNED_LONG_LONG to LONG_LONG. Hopefully this works with very old MPI installations.


cburstedde:
This might be the last change for version 2.8.6. Any comment @hannesbrandt @tim-griesbach ?

cburstedde:
Thanks, about to merge and release if there are no further comments.

cburstedde:
> > Thanks, about to merge and release if there are no further comments.
> 
> I saw that MPI_DOUBLE_INT is part of MPI standard 1.3. Do we want to remove sc_MPI_DOUBLE_INT nonetheless?

It's a weird thing, 12 bytes and all. But for the sake of backwards compatbility, letting it live and reducing the diff as much as possible would be a good thing. Would you be able to do this?

cburstedde:
> > > Thanks, about to merge and release if there are no further comments.
> > 
> > 
> > I saw that MPI_DOUBLE_INT is part of MPI standard 1.3. Do we want to remove sc_MPI_DOUBLE_INT nonetheless?
> 
> It's a weird thing, 12 bytes and all. But for the sake of backwards compatbility, letting it live and reducing the diff as much as possible would be a good thing. Would you be able to do this?

Don't worry, I'm on it.

PR #177, opened 2024-04-18
Build cmake examples by default

mkirilin:
This PR enables building examples by default and disables building and running them as tests.


cburstedde:
Thanks!

PR #178, opened 2024-04-25
Fix CMake variables

dutkalex:


cburstedde:
Thanks! Unifying the variables is great. Would you mind adding a file doc/author_<lastname>.txt like the others, releasing your code under the FreeBSD license? And if you could add a line to doc/release_notes.txt?

(Wondering why all of a sudden the autotools ./bootstrap script dies on Darwin (unrelated).)

Any comments @scivision -- thinking about callability from other software besides p4est.

dutkalex:
Done! @cburstedde

cmake/config.cmake
```diff
@@ -247,13 +235,13 @@ check_symbol_exists("SC_ENABLE_MPI" ${PROJECT_BINARY_DIR}/include/sc_config.h SC
 check_symbol_exists("SC_ENABLE_MPIIO" ${PROJECT_BINARY_DIR}/include/sc_config.h SC_ENABLE_MPIIO)
 
 # check consistency of MPI configuration
-if(mpi AND NOT (SC_ENABLE_MPI AND SC_ENABLE_MPIIO))
-  message(FATAL_ERROR "libsc MPI support was requested, but not configured in ${PROJECT_BINARY_DIR}/include/sc_config.h")
-endif()
-
-# check consistency of MPI I/O configuration
-if(mpi AND (SC_ENABLE_MPI AND NOT SC_ENABLE_MPIIO))
```
tim-griesbach, line 255: It may be useful to still check if `SC_ENABLE_MPI` is activated but `SC_ENABLE_MPIIO` not, since the latter may be not available.

tim-griesbach: We are checking this case explicitly to warn about the deprecated configuration setup.

tim-griesbach:
Thanks for the simplifications!

We tried out the changes and setting `SC_USE_INTERNAL_ZLIB` in `ccmake` to `ON` leads to the following configure error:
```
 CMake Error at cmake/zlib.cmake:60 (add_library):
   add_library cannot create imported target "ZLIB::ZLIB" because another
   target with the same name already exists.
 Call Stack (most recent call first):
   cmake/config.cmake:36 (include)
   CMakeLists.txt:15 (include)
```

cc @hannesbrandt

cmake/config.cmake
dutkalex: This is now checked in `check_mpiio.cmake`

tim-griesbach:
We saw that when one is initially calling `cmake` without any passed options and then calls `ccmake` to switch `SC_ENABLE_MPI` to `ON`, one gets the warning that MPI I/O is disabled/not found. However, an initial call of `cmake` with the command line parameter `-DSC_ENABLE_MPI=1` works fine. 

We tested also the code without the changes of this PR and copied the moved warnings to the same place as in this PR but the code still did not show the warning.

tim-griesbach:
> In the long run, it seems that namespace cleanliness would win, so we will go for the SC_ENABLE_* variables. Is there no way at all to have a short CMake setting via "mpi", maybe with presets, on the command/configure line?

After some further tinkering, we found a possible way. One could add
```
if (DEFINED mpi)
  set (SC_ENABLE_MPI ${mpi} CACHE BOOL "" FORCE)
endif()
```
directly before the first appearance of `SC_ENABLE_MPI` in `cmake/config.cmake`. While `mpi` does not show up in ccmake, it is still possible to use `-Dmpi`.

dutkalex:
> Thank you for the quick response! We tried out your recent changes and we get
> 
> ```
> -- Build files have been written to: /home/griesbac/workspace/sc_cmake/ZLIB-prefix/src/ZLIB-build
> [ 10%] Performing build step for 'ZLIB'
> [100%] Built target zlib
> [ 11%] Performing install step for 'ZLIB'
> [100%] Built target zlib
> Install the project...
> -- Install configuration: "Release"
> -- Installing: /usr/local/lib/libz.a
> CMake Error at cmake_install.cmake:46 (file):
>   file INSTALL cannot copy file
>   "/home/griesbac/workspace/sc_cmake/ZLIB-prefix/src/ZLIB-build/libz.a" to
>   "/usr/local/lib/libz.a": Permission denied.
> ```
> 
> Without the changes of this PR, we obtained
> 
> ```
> -- Build files have been written to: /home/griesbac/workspace/sc_cmake/ZLIB-prefix/src/ZLIB-build
> [ 10%] Performing build step for 'ZLIB'
> [100%] Built target zlib
> [ 11%] Performing install step for 'ZLIB'
> [100%] Built target zlib
> Install the project...
> -- Install configuration: "Release"
> -- Up-to-date: /home/griesbac/workspace/sc_cmake/local/lib/libz.a
> -- Up-to-date: /home/griesbac/workspace/sc_cmake/local/include/zlib.h
> -- Up-to-date: /home/griesbac/workspace/sc_cmake/local/include/zlib_name_mangling.h
> -- Up-to-date: /home/griesbac/workspace/sc_cmake/local/include/zconf.h
> -- Installing: /home/griesbac/workspace/sc_cmake/local/lib/pkgconfig/zlib.pc
> -- Up-to-date: /home/griesbac/workspace/sc_cmake/local/lib/libz.a
> -- Up-to-date: /home/griesbac/workspace/sc_cmake/local/lib/cmake/ZLIB/ZLIB.cmake
> -- Installing: /home/griesbac/workspace/sc_cmake/local/lib/cmake/ZLIB/ZLIB-release.cmake
> [ 13%] Completed 'ZLIB'
> ```
> 
> Apparently the installation directory of zlib changed to a path outside of the build directory.

Yes this is intentional. Hardcoding the install path into the CMakeLists.txt is a code smell. The install path is meant to be set on the command line as it depends on the target machine. Enforcing one breaks portability, composability with downstream software, and packaging tools such as CPack or Spack. In order to retrieve the old behavior you just need to specify it with a `-DCMAKE_INSTALL_PREFIX=...` or using the `--prefix` option of cmake.

dutkalex:
> We saw that when one is initially calling `cmake` without any passed options and then calls `ccmake` to switch `SC_ENABLE_MPI` to `ON`, one gets the warning that MPI I/O is disabled/not found. However, an initial call of `cmake` with the command line parameter `-DSC_ENABLE_MPI=1` works fine.
> 
> We tested also the code without the changes of this PR and copied the moved warnings to the same place as in this PR but the code still did not show the warning.

Very strange behavior. Maybe some cache issue...
Let me investigate and get back to you...

dutkalex:
> > In the long run, it seems that namespace cleanliness would win, so we will go for the SC_ENABLE_* variables. Is there no way at all to have a short CMake setting via "mpi", maybe with presets, on the command/configure line?
> 
> After some further tinkering, we found a possible way. One could add
> 
> ```
> if (DEFINED mpi)
>   set (SC_ENABLE_MPI ${mpi} CACHE BOOL "" FORCE)
> endif()
> ```
> 
> directly before the first appearance of `SC_ENABLE_MPI` in `cmake/config.cmake`. While `mpi` does not show up in ccmake, it is still possible to use `-Dmpi`.

Yes this seems like a good tradeoff: existing code would continue working, but new downstream projects would not notice it. I can include it in the PR if this is ok for everyone

dutkalex:
> Is there no way at all to have a short CMake setting via "mpi", maybe with presets, on the command/configure line?

> Hardcoding the install path into the CMakeLists.txt is a code smell.

Another solution to both of these problems could be to define a custom `SCDev` (or whatever) build mode which provides the desired defaults and shortcuts...

cmake/options.cmake
```diff
@@ -1,27 +1,13 @@
...
-option(BUILD_SHARED_LIBS "build shared libsc")
-option(CMAKE_TLS_VERIFY "verify TLS certificate" on)
+option( SC_USE_INTERNAL_ZLIB "build ZLIB" OFF )
+option( SC_USE_INTERNAL_JSON "build Jansson" OFF )
 
-# --- default install directory under build/local
-# users can specify like "cmake -B build -DCMAKE_INSTALL_PREFIX=~/mydir"
+option( SC_BUILD_SHARED_LIBS "build shared libsc" OFF )
+option( SC_BUILD_TESTING "build libsc self-tests" ON )
+option( SC_TEST_WITH_VALGRIND "run self-tests with valgrind" OFF )
```
elykwilliams, line 9: If changing this variable name, cmake/valgrind.cmake and test/CMakeLists.txt should be updated as well.

dutkalex: Thanks for letting me know. This should be fixed now

.github/workflows/ci_cmake.yml
```diff
@@ -133,7 +133,7 @@ jobs:
       uses: actions/checkout@v4
 
     - name: CMake configure
-      run: cmake --preset default -Dmpi:BOOL=${{ matrix.mpi }} --install-prefix=${{ runner.temp }} -DBUILD_SHARED_LIBS:BOOL=${{ matrix.shared }} -DTEST_WITH_VALGRIND:BOOL=${{ matrix.valgrind }}
+      run: cmake --preset default -DSC_ENABLE_MPI:BOOL=${{ matrix.mpi }} --install-prefix=${{ runner.temp }} -DSC_BUILD_SHARED_LIBS:BOOL=${{ matrix.shared }} -DTEST_WITH_VALGRIND:BOOL=${{ matrix.valgrind }}
```
elykwilliams, line 136: Test_with_valgrind should also be updated here

dutkalex: Done

dutkalex:
@cburstedde @tim-griesbach @elykwilliams Anything more you need to get this merged?

cburstedde:
Thanks so much! I'd go for the suggestion with `if (DEFINED mpi)`, and not relying on build modes.
Same with the `debug` configure variable.

Will this version be compatible with the present p4est, or do we need to update synchronously?

tim-griesbach:
> > We saw that when one is initially calling `cmake` without any passed options and then calls `ccmake` to switch `SC_ENABLE_MPI` to `ON`, one gets the warning that MPI I/O is disabled/not found. However, an initial call of `cmake` with the command line parameter `-DSC_ENABLE_MPI=1` works fine.
> > We tested also the code without the changes of this PR and copied the moved warnings to the same place as in this PR but the code still did not show the warning.
> 
> Very strange behavior. Maybe some cache issue... Let me investigate and get back to you...

@dutkalex  We still have this issue with the changes in this PR. Do you have an idea how to avoid this?

dutkalex:
Ah yes right! Thanks for reminding me! Last time I could not reproduce the issue but I'll give it another try

dutkalex:
> Thanks so much! I'd go for the suggestion with `if (DEFINED mpi)`, and not relying on build modes. Same with the `debug` configure variable.
> 
> Will this version be compatible with the present p4est, or do we need to update synchronously?

I think that it should work fine with the current state of p4est (develop) given that it already queries the prefixed variables:
`set(P4EST_ENABLE_MPI ${SC_ENABLE_MPI} CACHE BOOL "P4EST MPI support" FORCE)`

doc/author_dutka.txt
```diff
@@ -0,0 +1 @@
+I hereby place my contributions to libsc into the public domain under FreeBSD license.  Alexandre Dutka.
```
cburstedde, line 1: Thanks; it's best not to mention the public domain, it's unclear how that relates to BSD in the same context.

cburstedde:
Wondering about the CI in that one instance... and thanks for adding the mpi and debug shortcuts.

dutkalex:
> Ah yes right! Thanks for reminding me! Last time I could not reproduce the issue but I'll give it another try

I finally managed to replicate the problem. I don't quite understand what was the root cause, but it should be fixed now!

tim-griesbach:
> > Ah yes right! Thanks for reminding me! Last time I could not reproduce the issue but I'll give it another try
> 
> I finally managed to replicate the problem. I don't quite understand what was the root cause, but it should be fixed now!

Thank you for the fix! I tried out your changes and it works for me as well.

cburstedde:
> Wondering about the CI in that one instance... and thanks for adding the mpi and debug shortcuts.

This may be resolved after the latest merge into develop.

cburstedde:
So close. What's going on on Windows now (unrelated?)?

dutkalex:
I think so:
```
CMake Error at C:/Program Files/CMake/share/cmake-3.29/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find MPI (missing: MPI_C_FOUND C)
  ```

cburstedde:
> ```
> CMake Error at C:/Program Files/CMake/share/cmake-3.29/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
>   Could NOT find MPI (missing: MPI_C_FOUND C)
> ```

Thanks; I'd think @scivision would know how to repair this.

cburstedde:
> There was conflicting CMake variable use, fixed by this patch
> 
> [178.patch](https://github.com/cburstedde/libsc/files/15381488/178.patch)
> 
> Note that CMAKE_BUILD_TYPE must be set before `project()` or it has no effect.

Thanks; when I keep using -Dmpi the error persists. The most recent code has the purpose to allow for -Dmpi on the cmake command line while using SC_ENABLE_MPI as the variable that is used in the cmake code. Is there a way to make this work with `cmake -Dmpi`?

cburstedde:
Ok I've integrated it all with the latest develop version. Sorry for the merge in between. It got a bit confusing, but at least we're current with everything.

Would @dutkalex check whether this is all correct and I didn't screw up the merge?

dutkalex:
Looks good to me @cburstedde 😊

cburstedde:
Welcome to the developer list. :)

elykwilliams:
@dutkalex Can you please restore your branch, it seems one of your commits has broken the CI.

cburstedde:
> @dutkalex Can you please restore your branch, it seems one of your commits has broken the CI. 

It's a mix of disabling the OpenMPI support in Cmake and your code on the
SC_ variables.  I might have overlooked something in merging.  I do not know
what though -- do you think you can narrow it down (by bisect if need be)?

Sorry I cannot help until later next week, and thanks!  Feel free to reopen.

PR #181, opened 2024-05-09
CI MacOS needs libtool

scivision:
over time Github Actions runners change, and this fixes that change that needs libtool installed.


cburstedde:
Thanks o lot for fixing this! :)

PR #182, opened 2024-05-09
ci:cmake: use composite actions to deduplicate

scivision:
Composite actions avoid duplicating YaML. In coming weeks as ubuntu-24.04 runners become available I would add more compilers. This PR lays the groundwork to avoid duplicating so much YaML via composite actions.


cburstedde:
Cool! Keeping an eye open when we get to the CMake option variables for mpi and debug in #178.

Issue #179, opened 2024-05-04
libsc introduces warnings in downstream software


dutkalex:
**Description**
I don't get what the purpose of this macro is, but the compiler complains that it declares a function which is never called.

```
In file included from /Users/adutka/hopps/tpls/t8code/sc/src/sc.h:210:
/Users/adutka/hopps/tpls/t8code/sc/src/sc_mpi.h:804:1: warning: unused function 'sc_extern_c_hack_2' [-Wunused-function]
  804 | SC_EXTERN_C_END;
      | ^~~~~~~~~~~~~~~
/Users/adutka/hopps/tpls/t8code/sc/src/sc.h:200:51: note: expanded from macro 'SC_EXTERN_C_END'
  200 | #define SC_EXTERN_C_END                    } void sc_extern_c_hack_2 (void)
      |                                                   ^~~~~~~~~~~~~~~~~~
```

**To Reproduce**
Compile a dummy downstream project including `src/sc.h` with `-Wall` enabled (Clang 17)

cburstedde:
> I don't get what the purpose of this macro is, but the compiler complains that it declares a function which is never called.

Thanks!  This is a workaround to have a macro that hides the extern "C" { }
braces from GNU indent, for both C and C++.  We'll look into it!

cburstedde:
Thanks for the report, indeed the hack_2 function was subtly not extern "C". Does this resolve the warning?

cburstedde:
Resolved by #183.

PR #183, opened 2024-05-13
Feature extern c

cburstedde:
# Set C linkage for hack prototype 

Following up on issue #179.

Proposed changes: Make the prototype of sc_extern_c_hack_2 have C linkage, which allows to use the same names of the hack functions for both C and C++. Hopefully this avoids the warning with Clang. (Wondering why Clang would complain about declaring an unused prototype in a header, which is what headers do.)

For completeness, also implement the two dummy functions even though they are never called from the code.

### DIFF
--- a/doc/release_notes.txt
+++ b/doc/release_notes.txt
@@ -16,6 +16,10 @@
 
 ## Next release
 
+### Functionality
+
+ - Fix extern "C" macro for using C++ compilers with libsc .c files
+
 ### Build system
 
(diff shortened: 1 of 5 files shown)


cburstedde:
Pinging @dutkalex for checking this with clang-17, if you like.

dutkalex:
> Pinging @dutkalex for checking this with clang-17, if you like.

Looks good! No more warnings :)

PR #184, opened 2024-05-16
CI:CMake: add GCC-14

scivision:
Note: other PR using CMake may want to rebase on this PR once it's merged. This also fixes new failures in previously working CI due to changes in the past week in GitHub Actions runner images.

### DIFF
--- a/.github/workflows/ci_cmake.yml
+++ b/.github/workflows/ci_cmake.yml
@@ -21,27 +21,53 @@ env:
 
 jobs:
 
-  linux:
-    runs-on: ubuntu-22.04
-    name: CMake build on Linux
+  linux-gcc14:
+    runs-on: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


dutkalex:
> Cool! Pinging @dutkalex  wrt. the shared configure variable: is it handled by #178?

Can you elaborate? I'm not sure I quite understand your question @cburstedde

cburstedde:
> > Cool! Pinging @dutkalex  wrt. the shared configure variable: is it handled by #178?
> 
> Can you elaborate? I'm not sure I quite understand your question @cburstedde

I was thinking -Dmpi is becoming a shortcut to define SC_ENABLE_MPI in that PR, same for debug. With -Dshared, we do no such thing iirc, so this question may be void?

dutkalex:
> I was thinking -Dmpi is becoming a shortcut to define SC_ENABLE_MPI in that PR, same for debug. With -Dshared, we do no such thing iirc, so this question may be void?

I believe I did not remove any `-Dshared` flag (IIRC)

PR #186, opened 2024-05-26
Add markdown file extension

elykwilliams:
# Change file extension of README

Issue:
Currently, the README uses markdown formatting that is not recognized by github. 

Proposed changes:
By adding the .md extension to README, github will detect and format the README appropriately.


cburstedde:
Thanks. I'd be leaning towards maintaing a short README file with just the license and telling readers to read README.md for more details. So README.md would in this PR be a straight copy of the previous README.

Then we should make sure to distribute both files with both autoconf and CMake, and verify make distcheck and cpack.

cburstedde:
> Thanks. I'd be leaning towards maintaing a short README file with just the license and telling readers to read README.md for more details. So README.md would in this PR be a straight copy of the previous README.

The copying is done, this is cool. Adding a stub README would be good to have due to tradition/convention.

> Then we should make sure to distribute both files with both autoconf and CMake, and verify make distcheck and cpack.

For automake, adding README.md to the EXTRA_DIST variable would be required.

cburstedde:
So, to do this consistently across both p4est and libsc, it would be nice to keep README for GNU conformance, maybe just as a stub referencing a more extensive README.md. Plus we may take a step back and move some of the README contents into the doxygen pages, shortening README and adding various links. This requires a documentation bump on the web first.

Issue #187, opened 2024-05-26
Cmake: Failed to create symbolic link while installing


elykwilliams:
**Description**
I get the following error when specifying the install prefix path at install time to a non-default location
```
CMake Error: failed to create symbolic link '~/libsc/build/local/lib/pkgconfig/libsc.pc': No such file or directory
```

**To Reproduce**
```
git clone git@github.com:cburstedde/libsc.git
cd libsc
git checkout develop
mkdir build
mkdir install
cd build
cmake ../ .
cmake --build .
cmake --install . --prefix=../install
```

**Additional information**
cmake version 3.22.1

elykwilliams:
The issue seems to be that even though `${CMAKE_INSTALL_PREFIX}` is updated at install time, `install(CODE ... )` is using the cached value from configure time.

cburstedde:
> The issue seems to be that even though `${CMAKE_INSTALL_PREFIX}` is updated at install time, `install(CODE ... )` is using the cached value from configure time to try creating the symbolic link.
> 
> It looks like other people [have run into this](https://discourse.cmake.org/t/exit-with-error-if-user-attempt-to-override-cmake-install-prefix-during-installation/10020), and was addressed in [cmake 3.27](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:INSTALL_PREFIX) by allowing for `$<INSTALL_PREFIX>` to be evaluated in `install(CODE ... )`
> 
> While I see no problem in not supporting this for cmake versions less than 3.27, I'm not sure currently how one would detect and handle it gracefully.

Maybe @scivision has an idea of the CMake version checks and requirements.

cburstedde:
So something is definitely not right with the CMake CI, related to the most recent couple merges?

elykwilliams:
My comment here was largely irrelevant after I found the fix in the submitted PR. 




> So something is definitely not right with the CMake CI, related to the most recent couple merges?
Looking into it

cburstedde:
This issue may be resolved, since the CI fail relates to gcc-13 on MacOS?

PR #189, opened 2024-05-26
CMake: Fix pkgconf install path in symbolic link

elykwilliams:
# Fix pkgconf install path in symbolic link

Closes issue #187.

Proposed changes:
The `CMAKE_INSTALL_PATH` was being set an configure time and not using the value present at install time. This change will allow all supported cmake versions to specify the install prefix when installing via `cmake --install . --prefix=/path/to/install`. 

Note that this change will cause cmake to issue a fatal error and abort install if the symbolic link is not created, instead of failing passively as before.

### DIFF
--- a/cmake/pkgconf.cmake
+++ b/cmake/pkgconf.cmake
@@ -19,11 +19,8 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/${pc_filename}
         DESTINATION lib/pkgconfig)
 
 set(pc_target ${pc_filename})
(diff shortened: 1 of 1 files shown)


cburstedde:
Thanks! So far we had not given the pkgconfig support much coverage. If we were eventually to arrive at installed .pc files that actually work the way Linux distributions need it to, and are identical between autotools and CMake, that would be another major step forward.

cburstedde:
We may augment our CI eventually to check into the generated .pc contents.

Issue #188, opened 2024-05-26
Cmake: git-version-gen not found


elykwilliams:
**Description**
I noticed that when building using cmake I get the following file `libsc-.pc` which is supposed to have the git-version appended. I traced it back to `git-version-gen` not being found in `/build-aux`

**To Reproduce**
```
git clone git@github.com:cburstedde/libsc.git
cd libsc
git checkout develop
mkdir build
cd build
cmake ../ .
cmake --build .
cmake --install . 
ls local/lib/pkgconfig
```

**Additional information**
cmake version 3.22.1

PR #190, opened 2024-05-26
CMake: fix git-version-gen path

elykwilliams:
# Fix git-version-gen path

Closes issue #188.

Proposed changes:
find_program was previously using `${PROJECT_SOURCE_DIR}/build-aux` to look for `git-version-gen`, but because we use this before `PROJECT()` is called, it evaluates to empty causing `git-version-gen` to be not found. This is corrected to `${CMAKE_CURRENT_SOURCE_DIR}/build-aux`. 

This will fix the libsc version from being 0.0.0 when building from source, and fix the name of `libsc-{git-version}.pc`

Note that you will probably want to test that this works as expected when including libsc and subproject in p4est as per #167 .


cburstedde:
Thanks so much, it's important for us to have consistent version numbers.

Issue #191, opened 2024-05-31
CI: Valgrind test fails on Ubuntu 24.04


elykwilliams:
**Description**
When increasing the OS in from `ubuntu-22.04` to `ubuntu-24.04`, in the `linux-valgrind` CI, the `Install System Dependencies` fails with the error 
```
Restarting services...
 systemctl restart chrony.service containerd.service cron.service haveged.service hv-kvp-daemon.service multipathd.service packagekit.service php8.3-fpm.service polkit.service rsyslog.service runner-provisioner.service systemd-journald.service systemd-networkd.service systemd-resolved.service systemd-udevd.service udisks2.service walinuxagent.service
Error: The operation was canceled.
```
See https://github.com/cburstedde/libsc/actions/runs/9303838763 for details.

cburstedde:
Is this one resolved in the current CI on the develop branch?

Issue #185, opened 2024-05-16
Wondering about CI bitrot


cburstedde:
The CI fails occasionally on github workflow updates. @scivision addresses these issues on a regular basis; thanks!

Considering our focus on long-term support: Can we do something to make our CI framework more futureproof and robust against whatever the github CI system may come up with? Is there a way to be long-term compatible with a minimum amount of maintenance?

cburstedde:
Sounds reasonable! Let's just keep using those CI features that are most promising to be long term stable.

cburstedde:
I'm keeping this one open. The goal is to stablilize the CI such that by the end of the year it can live on its own without further maintenance. To this end, we may eventually need to go from ubuntu versions to latest.

cburstedde:
> I'm keeping this one open. The goal is to stablilize the CI such that by the end of the year it can live on its own without further maintenance. To this end, we may eventually need to go from ubuntu versions to latest.

I'd like to stay away from bleeding-edge build tool versions in the CI. To be as maintenance free as possible, I'd like to propose to use no newer packages than what's available in Debian stable (unless they are the default of ubuntu-latest or some other generic OS option). Ideally I'd like to stop using version numbers altogether to avoid technical debt.

cburstedde:
If we use github actions with version tags like checkout@v4, they will run out of life at some point and github starts deprecating them. It seems using @main instead solves the problem indefinitely -- as long as github does not change the main branch name.

cburstedde:
sc is still doing fine! :) I was provoking some bugs on the non-MPI build, which seems to be CI'ed from the autotools builds but currently not from CMake. While this build is rarely relevant, it would be nice to update the CMake CI with one non-MPI build, and if @scivision likes the idea to go for @main in the actions.

PR #192, opened 2024-05-31
CI: Revert breaking change to valgrind test

elykwilliams:
# Revert breaking change to valgrind CI test

It looks like the Valgrind test will need some updating before bumping the OS to Ubuntu 24.04. I'm not sure why but as the logs show, it's failing to successfully install the system packages.  Related to #185. Breaking Commit was 0c3a681.

Proposed changes:
I've reverted the OS to Ubuntu 22.04 for the valgrind test, until I can give it a closer look #191 .

### DIFF
--- a/.github/workflows/ci_cmake.yml
+++ b/.github/workflows/ci_cmake.yml
@@ -89,7 +89,7 @@ jobs:
 
   linux-valgrind:
     needs: linux-gcc14
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-22.04
     name: Valgrind Linux mpi=${{ matrix.mpi }} CC=${{ matrix.cc }} shared=${{ matrix.shared }}
     timeout-minutes: 60
 


cburstedde:
Thanks, is there anything we can do about the faliing test on Mac with gcc 13?

elykwilliams:
> Thanks, is there anything we can do about the faliing test on Mac with gcc 13?

I'm looking into it. It's strange because the bump to gcc-13 happened back in December, so that by itself isnt the problem. But the issue seems to be that cmake isn't finding a compatible version of mpi, even though it is finding the homebrew installed open-mpi.

cburstedde:
Maybe @scivision can help.

As soon as possible, I'd like to establish to only use software (like valgrind and CMake) no newer than what's in the current Debian stable distribution. For CMake that's 3.25.

cburstedde:
Should I just merge this one since it CI checks fine?

cburstedde:
> Maybe @scivision can help.
> 
> As soon as possible, I'd like to establish to only use software (like valgrind and CMake) no newer than what's in the current Debian stable distribution. For CMake that's 3.25.

That would also mean using at most gcc-11. Would anything speak against this policy, specifically for p4est and sc?

cburstedde:
I'm also wondering about specifying CC= in the CI CMake. With MPI, the compiler should always be mpicc -- is this silently added by the CMake MPI logic, or do we need to do something about it?

cburstedde:
> I think this PR can be closed; this was a transient issue in the new Ubuntu 24.04 image. #194 shows Valgrind working with Ubuntu 24.04

Ok closing here. Thanks for the report/initiative @elykwilliams !

PR #156, opened 2024-01-08
Update scda error management

tim-griesbach:
# Update scda error management

Proposed changes: Change the scda error management and the corresponding documentation. In concrete terms, this means
- write a general error management paragraph in the doxygen documentation,
- add the API of the function `sc_scda_ferror_class` and
- introduce the struct `sc_scda_ferror_t` and document its semantic.

PR #132, opened 2023-07-18
Add scda API

tim-griesbach:
# Add scda API

This PR introduces in libsc the API of the scda file format (https://doi.org/10.48550/arXiv.2307.06789), a file-oriented data format suitable for parallel, partition-independent disk I/O. The main purpose of the format is to abstract any parallelism and provide sufficient structure as a foundation for a generic and flexible archival and checkpoint/restart.

We only add the API for now to gather feedback on the API and also on the format as introduced in the arxiv preprint linked above. We appreciate any feedback on scda and its concrete API in libsc.

Proposed changes:
- Add  the API of the scda file format in `sc_scda.h`
- Restructure IO module in the Doxygen documentation and introduce a page for the scda workflow
- General polishing and completion of the documentation in `sc_io.h`

PR #193, opened 2024-06-12
scda: Write and read a file header

tim-griesbach:
# scda: Write and read a file header

This PR adds the functionality to read and write a scda file header, i.e. implementing in particular the functions `fopen_write` and `fopen_read` according to the scda API introduced in https://github.com/cburstedde/libsc/pull/132 and https://github.com/cburstedde/libsc/pull/156.

A list of some of the functionalities that are implemented in this PR is
- `fopen_read` and `fopen_write`,
- helper functions that pad data according to the scda convention defined in the [scda preprint]( 	
https://doi.org/10.48550/arXiv.2307.06789),
- helper functions that check if data is correctly padded and that extract the actual data,
- complete error handling for `fopen_read` and `fopen_write`; including MPI count error management,
- `ferror_string` to convert error codes to error strings,
-  fuzzy error checking to increase the code coverage of the error management,
- a basic test program that currently aborts on errors and offers fuzzy error testing by command line parameters and
- the non-static function `sc_scda_is_success` to check if an error code encodes success.


cburstedde:
Would this CI check ok when moving gcc back to -11 on MacOS?

tim-griesbach:
> Would this CI check ok when moving gcc back to -11 on MacOS?

I think so since the CI failure in this PR is the same as in the PR https://github.com/cburstedde/libsc/pull/192, which has disjoint code changes. Furthermore, the scda PR implements scda functions and is not changing the way CMake finds MPI. So, I do not see a connection to the CI error.

cburstedde:
When merging the latest develop branch, will the CI go through?

tim-griesbach:
> When merging the latest develop branch, will the CI go through?

Yes, I think so. I merged the latest develop branch into this PR and the CI looks good to me.

src/sc_scda.h
```diff
@@ -1045,7 +1068,15 @@ sc_scda_fcontext_t *sc_scda_fread_varray_data (sc_scda_fcontext_t * fc,
...
-                                          sc_scda_ferror_t *errclass);
+                                          sc_scda_ferror_t * errclass);
+
+/** Check if a scda_errorcode_t encodes success.
+ *
+ * \param [in]    errorcode     An errcode that is output by a sc_scda function.
+ * \return                      True if \b errcode encodes success and
+ *                              false otherwise.
+ */
+int                 sc_scda_is_success (sc_scda_ferror_t errorcode);
```
cburstedde, line 1079: Thanks! I'll be merging this one now.

For future development: Would it be more intuitive to rename the function to sc_scda_ferror_success (or something else -- open for suggestions) and to return MPI_SUCCESS on success and some MPI error otherwise?

tim-griesbach: I think that a function `sc_scda_ferror_success` that returns MPI error codes may be unexpected to the user and we also have the `sc_scda_ret_t` value `SC_SCDA_FERR_SUCCESS` that may be named too similar.
Do you think it is sensible to name the function `sc_scda_ferror_is_success` to be consistent to the other scda function names but still use the semantic of mapping an error code to a Boolean?

cburstedde:
> +/** Check if a scda_errorcode_t encodes success.
> + *
> + * \param [in]    errorcode     An errcode that is output by a sc_scda function.
> + * \return                      True if \b errcode encodes success and
> + *                              false otherwise.
> + */
> +int                 sc_scda_is_success (sc_scda_ferror_t errorcode);
> 
> I think that a function `sc_scda_ferror_success` that returns MPI error codes may be unexpected to the user and we also have the `sc_scda_ret_t` value `SC_SCDA_FERR_SUCCESS` that may be named too similar.

Good point.

> Do you think it is sensible to name the function `sc_scda_ferror_is_success` to be consistent to the other scda function names but still use the semantic of mapping an error code to a Boolean?

I like this proposal!

PR #194, opened 2024-06-20
ci:cmake: use default GCC,Clang per OS

scivision:
per Carsten, simplify compiler fanout by using only default GCC or Clang.

### DIFF
--- a/.github/workflows/ci_cmake.yml
+++ b/.github/workflows/ci_cmake.yml
@@ -18,26 +18,26 @@ env:
   CTEST_NO_TESTS_ACTION: "error"
   CMAKE_BUILD_PARALLEL_LEVEL: 4
   CTEST_PARALLEL_LEVEL: 0
+  CMAKE_INSTALL_PREFIX: ~/local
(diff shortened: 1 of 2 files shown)


cburstedde:
Thanks; how can we make the CI go through?

.github/workflows/ci_cmake.yml
```diff
@@ -18,27 +18,27 @@ env:
   CTEST_NO_TESTS_ACTION: "error"
   CMAKE_BUILD_PARALLEL_LEVEL: 4
   CTEST_PARALLEL_LEVEL: 0
+  CMAKE_INSTALL_PREFIX: ~/libs
```
cburstedde, line 21: In p4est we default the install to `local` relative to the local directory. Would it make sense to align this?

cburstedde:
> Thanks; how can we make the CI go through?

Wondering what's going on on Windows, if this is a regression or a CI feature.

cburstedde:
Cool, trying this out!

PR #197, opened 2024-07-16
scda: Parallel handling of non-collective errors without `goto`

tim-griesbach:
# scda: Parallel handling of non-collective errors without `goto`

This PR moves the serial code in the functions `fopen_{write,read}` to separate functions to make it possible to jump out of the serial code part in case of an error without using `goto`. The error macros and their documentation are adjusted accordingly.
Moreover, this PR renames `sc_scda_is_success` to `sc_scda_ferror_is_success`, makes the padded data argument of the padding extraction functions `const` and adds documentation for the function `sc_scda_check_file_header`.

### DIFF
--- a/src/sc_scda.c
+++ b/src/sc_scda.c
@@ -284,8 +284,8 @@ sc_scda_pad_to_fix_len (const char *input_data, size_t input_len,
  *                            False, otherwise.
  */
 static int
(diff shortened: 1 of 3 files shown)


cburstedde:
Thanks; I like the modularization! For continued development, it migth make sense to introduce even more, smaller helper functions.

PR #202, opened 2024-09-17
scda: Fuzzy error synchronization and padding optimization

tim-griesbach:
# scda: Fuzzy error synchronization and padding optimization

This PR continues the implementation of the scda file format (cf. https://github.com/cburstedde/libsc/pull/201) by adding two smaller changes.
1. Parallel synchronization of the state of the random number generator for the fuzzy error testing. This is required since the serial code paths cause differing counts of drawn random samples on each MPI rank. This means that the fuzzy error return could make a collective error return to a non-collective error return, which is invalid behavior. Therefore, the state of the random number generator is synchronized after every non-collective code section. This change only impacts the code when fuzzy error return is enabled, leaving the normal scda code behavior unchanged.
2. Introduce the function `sc_scda_check_pad_to_fix_len` to check fixed-length padding without extracting the raw data. This function enables more modularization, saves one unnecessary copy of the vendor string and makes the set of padding functions for both padding options symmetric.

### DIFF
--- a/src/sc_scda.c
+++ b/src/sc_scda.c
@@ -131,6 +131,7 @@
                                     SC_CHECK_MPI(sc_MPI_Bcast(&errcode->mpiret,\
                                                   1, sc_MPI_INT, root,         \
                                                   fc->mpicomm));               \
+                                    sc_scda_fuzzy_sync_state (fc);             \
                                     if (!sc_scda_ferror_is_success (*errcode)) {\
                                     sc_scda_file_error_cleanup (&fc->file);    \
                                     SC_FREE (fc);                              \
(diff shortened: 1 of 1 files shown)


cburstedde:
Sounds good! Do we have a way to indicate in the code very clearly which parts are collective and which are not? This seems very important for the parallel consistency of error checking. Maybe by means of extra function calls into non-collective paths with proper naming? "Just" comments and conventions may not be sufficiently safe.

tim-griesbach:
The non-collective code paths are already in separate functions and the current naming convention for these static functions is that their name ends with `_internal`. I could replace the current name endings by `_serial` to be more clear on the functions' purpose.

PR #203, opened 2024-09-19
CMake build examples flag

dutkalex:
Separate the building of the tests from the building of the examples.
For reference: https://github.com/cburstedde/p4est/pull/316

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,12 +69,11 @@ add_subdirectory(src)
 
 if(SC_BUILD_TESTING)
   include(cmake/valgrind.cmake)
-
-  if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.22)
-    add_subdirectory(example)
-  endif()
-
   add_subdirectory(test)
+endif()
+
+if(SC_BUILD_EXAMPLES AND CMAKE_VERSION VERSION_GREATER_EQUAL 3.22)
+  add_subdirectory(example)
 endif()
 
 # --- packaging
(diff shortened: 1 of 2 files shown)


cburstedde:
Cool; I'm wondering what's up with the CI?

cburstedde:
Would you know what we need the CMake version condition for?

dutkalex:
> Cool; I'm wondering what's up with the CI?

It doesn't seem related to the CMake change, since the autotools build also fails...

> Would you know what we need the CMake version condition for?

No idea. It was there so I left it as is, but I really don't know what was the rationale behind

cburstedde:
> > Cool; I'm wondering what's up with the CI?
> 
> It doesn't seem related to the CMake change, since the autotools build also fails...

So we should have a look at the current state of the CI on the develop
branch of libsc?  Maybe @mkirilin has an idea.

> > Would you know what we need the CMake version condition for?
> 
> No idea. It was there so I left it as is, but I really don't know what was the rationale behind

It would be interesting to check the CMake example build code, maybe we can
remove it.  @scivision may know what it is needed for?

dutkalex:
Spot-on @mkirilin. That was indeed the issue. Thanks for your help 😊

CMakeLists.txt
```diff
@@ -62,14 +62,13 @@ add_subdirectory(src)
 
 if(SC_BUILD_TESTING)
   include(cmake/valgrind.cmake)
-
-  if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.22)
-    add_subdirectory(example)
-  endif()
-
```
cburstedde, line 69: I like it that the new code is out of the TESTING if-branch.

cburstedde:
Thanks for figuring it out! We can follow up on the version tests separately.

PR #207, opened 2024-11-11
Fix variable change check in CMake

tim-griesbach:
# Fix variable change check in CMake

### Problem description
Reconfiguring with enabled MPI after starting from a CMake configuration with MPI disabled results in a setup with enabled MPI but deactivated  MPI I/O, which is deprecated. The expected behavior is that MPI I/O is enabled if MPI is enabled.

### Proposed changes
The issue arises because `$CACHE{SC_ENABLE_MPI}` reflects the current MPI setting rather than the previous one. As a result, the condition
```
NOT "${SC_ENABLE_MPI}" STREQUAL "$CACHE{SC_ENABLE_MPI}"
```
evaluates to false, preventing the check for MPI I/O when MPI was previously disabled. To address this, I introduced a new variable, `CACHED_SC_ENABLE_MPI`, to track the last MPI setting, ensuring that the last used value of `SC_ENABLE_MPI` is preserved.

### DIFF
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -98,11 +98,13 @@ set(SC_ENABLE_PTHREAD ${CMAKE_USE_PTHREADS_INIT})
 set(SC_ENABLE_MEMALIGN 1)
 
 # user has requested a different MPI setting, so we need to clear these cache variables to recheck
-if(NOT "${SC_ENABLE_MPI}" STREQUAL "$CACHE{SC_ENABLE_MPI}")
+if(NOT "${SC_ENABLE_MPI}" STREQUAL "${CACHED_SC_ENABLE_MPI}")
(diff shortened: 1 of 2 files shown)


cburstedde:
Thanks! Are we sure not to miss out on any cache-related automatisms that CMake would invoke automatically with the older code?

tim-griesbach:
> Thanks! Are we sure not to miss out on any cache-related automatisms that CMake would invoke automatically with the older code?

To my understanding, a change of the cache variable `SC_ENABLE_MPI` using `ccmake` or using a parameter for the `cmake` command, results in a change of `CMakeCache.txt` and hence of `$CACHE{SC_ENABLE_MPI}` as well. I could not find any options/mechanisms that would change this behavior.

PR #208, opened 2024-11-12
scda: Custom MPI data type for indirect arrays

tim-griesbach:
# scda: Custom MPI data type for indirect arrays

### scda changes
This PR continues the scda implementation (cf. https://github.com/cburstedde/libsc/pull/205) by implementing and using a custom MPI data type to write and read indirect arrays without copying the local data to a contiguous buffer. This optimization is only performed if MPI is available and enabled. Moreover, `sc_scda_fopen_{write,read}` are changed to abort on the deprecated configuration of MPI being enabled but MPI I/O being disabled. This is also explained in the documentation and the abort message.

### Build system changes
The creation of the custom MPI data type in the function `sc_scda_get_indirect_type` requires the MPI functions `MPI_Get_address` and `MPI_Type_create_struct` and the type `MPI_Aint`, which are all part of the MPI 2.0 standard.  Since the scda functions are only valid to call if MPI I/O is available or no MPI at all is activated (cf. above), we know that MPI 2.0 must be supported. Without MPI `sc_scda_get_indirect_type` is not compiled and the internal contiguous buffer is used.

The function `sc_scda_get_indirect_type` also requires the MPI function `MPI_Aint_diff`. However, this is a MPI 3.0 function. Therefore, we check in Autoconf and CMake if the function is available. If the function is not available, we use a fallback that conforms to an example from the MPI 2.0 standard document (cf. page 283 in https://www.mpi-forum.org/docs/mpi-2.0/mpi2-report.pdf), i.e. standard subtraction of MPI_Aint. The user only needs to call the new function `sc_MPI_Aint_diff` that abstracts the described logic and the function also works without MPI. The introduction of `sc_MPI_Aint_diff` includes the introduction of `sc_MPI_Aint` in the same vein as the other wrapped MPI types.

### DIFF
--- a/cmake/check_mpiaintdiff.cmake
+++ b/cmake/check_mpiaintdiff.cmake
@@ -1,0 +1,23 @@
+include(CheckCSourceCompiles)
+function(check_mpiaintdiff)
+
+  set(CMAKE_REQUIRED_LIBRARIES MPI::MPI_C)
+
+  check_c_source_compiles(
+    "
+        #include <mpi.h>
+        int main(void) {
+          MPI_Aint a, b, res;
+          a = 42;
+          b = 12;
+          MPI_Init ((int *) 0, (char ***) 0);
+          res = MPI_Aint_diff (a, b);
+          MPI_Finalize ();
+          return 0;
+        }
+    "
+    SC_HAVE_AINT_DIFF)
+
+endfunction()
+
+check_mpiaintdiff()
--- a/cmake/check_mpiio.cmake
+++ b/cmake/check_mpiio.cmake
@@ -42,11 +42,6 @@ function(check_mpiio)
     "
     SC_ENABLE_MPIIO)
 
-    if( NOT SC_ENABLE_MPIIO )
-        message(WARNING "libsc MPI configured but MPI I/O is not configured/found: DEPRECATED")
-        message(NOTICE "This configuration is DEPRECATED and will be disallowed in the future.")
-        message(NOTICE "If the MPI File API is not available, please disable MPI altogether.")
-    endif()
 endfunction()
 
 check_mpiio()
--- a/cmake/config.cmake
+++ b/cmake/config.cmake
@@ -234,3 +234,12 @@ set(CMAKE_REQUIRED_DEFINITIONS)
 # libsc and current project must both be compiled with/without MPI
 check_symbol_exists("SC_ENABLE_MPI" ${PROJECT_BINARY_DIR}/include/sc_config.h SC_ENABLE_MPI)
 check_symbol_exists("SC_ENABLE_MPIIO" ${PROJECT_BINARY_DIR}/include/sc_config.h SC_ENABLE_MPIIO)
+
+# Check for deprecated MPI and MPI I/O configuration.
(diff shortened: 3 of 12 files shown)


cburstedde:
> This PR continues the scda implementation (cf. https://github.com/cburstedde/libsc/pull/205) by implementing and using a custom MPI data type to write and read indirect arrays without copying the local data to a contiguous buffer. This optimization is only performed if MPI is available and enabled. Moreover, `sc_scda_fopen_{write,read}` are changed to abort on the deprecated configuration of MPI being enabled but MPI I/O being disabled. This is also explained in the documentation and the abort message.

Sounds good!  In addition, are we warning loudly enough about this
configuration already when configure and CMake are executed?

> The creation of the custom MPI data type in the function `sc_scda_get_indirect_type` requires the MPI functions `MPI_Get_address` and `MPI_Type_create_struct` and the type `MPI_Aint`, which are all part of the MPI 2.0 standard.  Since the scda functions are only valid to call if MPI I/O is available or no MPI at all is activated (cf. above), we know that MPI 2.0 must be supported. Without MPI `sc_scda_get_indirect_type` is not compiled and the internal contiguous buffer is used.

Cool, so there is nothing to check.

> The function `sc_scda_get_indirect_type` also requires the MPI function `MPI_Aint_diff`. However, this is a MPI 3.0 function. Therefore, we check in Autoconf and CMake if the function is available. If the function is not available, we use a fallback that conforms to an example from the MPI 2.0 standard document (cf. page 283 in https://www.mpi-forum.org/docs/mpi-2.0/mpi2-report.pdf), i.e. standard subtraction of MPI_Aint. The user only needs to call the new function `sc_MPI_Aint_diff` that abstracts the described logic and the function also works without MPI. The introduction of `sc_MPI_Aint_diff` includes the introduction of `sc_MPI_Aint` in the same vein as the other wrapped MPI types.

Here I'd be wondering: what about the simpler alternative of not querying
these MPI 3 functions at all and always using the fallback?  Pros/Cons?

tim-griesbach:
> Sounds good! In addition, are we warning loudly enough about this configuration already when configure and CMake are executed?

I tested both again and I think both build systems warn loudly by printing a warning during the configuration that contains in both cases "DEPRECATED". However, the difference between the two build systems is that the Autconf-based configure prints the warning at the end of the configuration output but CMake prints the warning when MPI is checked. Hence, it may be necessary to scroll to see the warning in CMake. Therefore, I would move the warning in CMake to the end of configure output as for the Autconf-based configure. 

> Here I'd be wondering: what about the simpler alternative of not querying these MPI 3 functions at all and always using the fallback? Pros/Cons?

According to [this mpi-forum discussion](https://github.com/mpi-forum/mpi-forum-historic/issues/349), `MPI_Aint_diff` was added to ensure portable address arithmetic:
```
The MPI_AINT_ADD and MPI_AINT_DIFF routines were added to provide a portable
mechanism for manipulating addresses stored in MPI address integer types.
``` 
Therefore, I think always using the fallback should work on standard systems/setup and would result in a simplification but we would loose portability.

cburstedde:
Thanks again for the contribution!

PR #210, opened 2024-12-05
Fix fields types for sc3_array_t 

mkirilin:
Fix fields types for sc3_array_t 

Proposed changes:

This PR fixes the ,,sc3_array_t::ecount`` and ,,sc3_array_t::ealloc`` types from ,,int`` to ,,size_t`` as it is in sc_array_t. It will be beneficial for solving some type conflicts in shared memory partition and iteration algorithms.


cburstedde:
Wondering, why do we have 76 changed files? Is there no way to have the origin and new branch receive the same merges beforehand?

mkirilin:
I tried to push my feature-sc3 with merged the latest develop branch into yours feature-sc3. It caused the significant diff. Now I undid the latest develop merge.

I suppose, it caused the CI fail with MacOS, as I think this issue was solved on the develop branch.

The one suggestion would be to update ours both feature-sc3 branches with the newest develop.

cburstedde:
Thanks! Yes, updating both branches will make sense. I'll do this once I've released the next version of sc.

Using size_t consistently for memory sizes/locations makes sense. I would just suggest to rename a standard int variable like 'i' to 'iz' or similar to avoid any risk of confusion. ecount etc. can remain as is.

mkirilin:
> Using size_t consistently for memory sizes/locations makes sense. I would just suggest to rename a standard int variable like 'i' to 'iz' or similar to avoid any risk of confusion. ecount etc. can remain as is.

Thanks for the tip! Fixed.
#
Additionally, I pushed a small fix for checking `SC_ENABLE_OMPICOMMSOCKET` variable in cmake.
https://github.com/cburstedde/libsc/blob/6ee6679b65593fd512e8142f905fa32879d33d49/cmake/config.cmake#L112

It's a temporary solution, later I'll make a PR on refactoring it.

mkirilin:
Update: I pushed changes that fix different environmental variable name for Autoconf and CMake by changing `.._ENABLE_MPICOMMSOCKET` to `.._ENABLE_OMPICOMMSOCKET`.

Issue #206, opened 2024-10-31
Libsc failes to compile with CMake on MinGW


jmark:
**Description**

Compilation failes with CMake on MinGW platform. This problem is associated with creating
a t8code JLL package. See this pull request: https://github.com/JuliaPackaging/Yggdrasil/pull/9718

The error output comes at llinking libsc with:
`cannot find $<1:wsock32` and `cannot find -lIphlpapi`.
Especially, the first error output suggest that the generator expression is somehow wrong.

The used CMake version is `v3.21`.

The following patch solved the issue:
[T/t8code/bundled/patches/libsc.patch](https://github.com/JuliaPackaging/Yggdrasil/pull/9718/files#diff-bf9534cb6a089df4b2c07a921fffc6504712244753049a4582f69db51866053c)

But it is not clear to me if this is a proper solution.

**To Reproduce**

For reproduction, I suggest to try to create a t8code JLL package via Yggdrasil without the application of the patches.

```
cd T/t8code/
julia build_tarballs.jl --debug=error --verbose x86_64-w64-mingw32-mpi+microsoftmpi
```

**Additional information**

I used Julia 1.11. But this shouldn't have any effect.

cburstedde:
Thanks. I wouldn't mind adding a SC_WIN32 variable (with the prefix). @scivision: would that conflict with our windows builds that you usually execute?

jmark:
Thank you all for your input! I consider this issue resolved.

PR #212, opened 2024-12-17
Update cmake buildsystem

Davknapp:
# Update CMake-build system.

Following up on issue #206

Proposed changes: This PR cleans the sc-targets.cmake file, if the generator expression fails to be resolved. Furthermore it improves readability.


cburstedde:
Thanks; and welcome to the list of developers! Running CI checks.

Issue #213, opened 2024-12-28
Transient CI failures


cburstedde:
There are rather frequent transient CI failures. Even rerunning several times resolves them only slowly. I'm wondering if there's anything in our powers to reduce or eleminate these. Are we sure that there are no race conditions involving file access?

tim-griesbach:
> Is there a particular workflow (CI for CMake, CI for Autotools) that is failing intermittently? Is it a specific test (like 'scda') or subset of tests that's failing?

Based on my observations, it is only the `scda` test that causes transient CI failures. The CI failures seem to be due to unexpected MPI count values retrieved from MPI staus objects, i.e. a MPI I/O function completes without an error but reports a lower count of written/read bytes. This causes an scda count error (`SC_SCDA_FERR_COUNT`) and hence the CI fails. 

> Are we sure that there are no race conditions involving file access?

I didn't spot anything along these lines on the level of  MPI_File calls.

In p4est, we have a similar situation in `test_loadsave{2,3}`, e.g. https://github.com/cburstedde/p4est/actions/runs/12528069581. These tests use `p4est_{load,save}`, which perform MPI I/O directly through MPI_File functions, bypassing the `sc_io` file function wrappers for legacy reasons. Despite this difference, I observed analogous behavior: no MPI I/O errors, yet occasional incorrect MPI I/O counts during the CI runs. To sum it up, we have a substantially different MPI I/O workflow in p4est's `test_loadsave{2,3}` that leads to very similar transient CI fails as for scda.

In summary, it seems likely to me that the file system in GitHub Actions is contributing to the transient CI failures. One approach to mitigate the issue could be to implement a loop in the respective `sc_io` wrapper functions to repeat the I/O call until the sum of reported counts matches the expected count. We could introduce a maximal number of loop passes and maybe there is a reasonable number of maximal loop passes to avoid the described issue.

cburstedde:
Thanks all for the analysis.

> > Is there a particular workflow (CI for CMake, CI for Autotools) that is failing intermittently? Is it a specific test (like 'scda') or subset of tests that's failing?
> 
> Based on my observations, it is only the `scda` test that causes transient CI failures. The CI failures seem to be due to unexpected MPI count values retrieved from MPI staus objects, i.e. a MPI I/O function completes without an error but reports a lower count of written/read bytes. This causes an scda count error (`SC_SCDA_FERR_COUNT`) and hence the CI fails. 
> 
[...]

PR #214, opened 2025-01-15
Patch transient CI failures due to busy file system

tim-griesbach:
# Patch transient CI failures due to busy file system

Following up on issue https://github.com/cburstedde/libsc/issues/213.

As discussed in the referenced issue, we observed transient CI failures due to `test_scda.c`. The failures appear due to I/O operations that are according to MPI I/O's error code successful but have an unexpected MPI count of 0, indicating that no actual I/O operation was performed. During the investigation of this issue, I observed that calling a sleep between I/O operations reduces the frequency of the transient failures or even seems to overcome them, suggesting that the underlying cause may be a busy file system on the CI runners.

This PR applies a patch in the CI to wait 5 milliseconds if a successful I/O operation unexpectedly resulted in count 0 and then retry the I/O operation once. The sleeping time can be adjusted in the CI by adjusting the variable `IO_SLEEP_TIME` (cf. eb190eb072c0c95222898412e596d5aff6cedffa). In my tests 5 milliseconds ensured a reliable CI. However, past experiences indicate that the extent of the issue can vary significantly over time. Hence, it may be necessary to adjust the choice of the sleeping time in the future.

Edit: This PR contains now a different approach of avoiding the described issue (cf. https://github.com/cburstedde/libsc/pull/214#issuecomment-2639344963).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,8 +42,8 @@ jobs:
       run: |
         DIR="checkMPIdebug_shared" && mkdir -p "$DIR" && cd "$DIR"
         ../configure --enable-mpi --enable-debug --disable-shared \
-            CFLAGS="-O0 -g -Wall -Wextra -Wno-unused-parameter" \
-            $SC_CI_CONFIG_OPT
+                     CFLAGS="-O0 -g -Wall -Wextra -Wno-unused-parameter" \
+                     $SC_CI_CONFIG_OPT
         make -j V=0
         make -j check V=0
 
@@ -60,7 +60,7 @@ jobs:
       run: |
         DIR="checkdebug" && mkdir -p "$DIR" && cd "$DIR"
         ../configure --enable-debug CFLAGS="-O0 -g -Wall -Wno-uninitialized"\
-            $SC_CI_CONFIG_OPT
+                     $SC_CI_CONFIG_OPT
         make -j V=0
         make -j check V=0
 
@@ -74,7 +74,7 @@ jobs:
       run: |
         DIR="checkMPIdebugCXX" && mkdir -p "$DIR" && cd "$DIR"
         ../configure --enable-mpi --enable-debug CFLAGS="-O0" CC=mpicxx \
-            $SC_CI_CONFIG_OPT
+                     $SC_CI_CONFIG_OPT
         make -j V=0
         make -j check V=0
 
@@ -83,7 +83,7 @@ jobs:
       run: |
         DIR="checkOpenMPMPIdebug" && mkdir -p "$DIR" && cd "$DIR"
         ../configure --enable-openmp="-fopenmp" --enable-mpi \
-            --enable-debug CFLAGS="-O0" $SC_CI_CONFIG_OPT
+                     --enable-debug CFLAGS="-O0" $SC_CI_CONFIG_OPT
         make -j V=0
         make -j check V=0
 
@@ -127,8 +127,8 @@ jobs:
         DIR="checkMPIdebug_valgrind" && mkdir -p "$DIR" && cd "$DIR"
         ../configure --enable-mpi --enable-debug \
                      --disable-shared --enable-valgrind \
-          CFLAGS="-O0 -g -Wall -Wextra -Wno-unused-parameter" \
-          $SC_CI_CONFIG_OPT
+                     CFLAGS="-O0 -g -Wall -Wextra -Wno-unused-parameter" \
+                     $SC_CI_CONFIG_OPT
         make -j V=0
         make -j check V=0
 
@@ -172,9 +172,9 @@ jobs:
       run: |
         DIR="tarball" && mkdir -p "$DIR" && cd "$DIR"
         ../configure --enable-mpi --enable-debug \
-            CFLAGS="-O0 -g -pedantic -Wall -Wextra -Werror \
-                -Wno-unused-parameter -Wno-builtin-declaration-mismatch \
-                -Wno-implicit-fallthrough" $SC_CI_CONFIG_OPT
+                     CFLAGS="-O0 -g -pedantic -Wall -Wextra -Werror \
+                     -Wno-unused-parameter -Wno-builtin-declaration-mismatch \
+                     -Wno-implicit-fallthrough" $SC_CI_CONFIG_OPT
         make -j V=0
         make -j check V=0
         make -j distcheck V=0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,11 @@ on:
       - ".github/workflows/composite-cmake/**"
   pull_request:
 
+env:
+  # Due to busy file system problems on the CI runners, we deactivate the file
+  # checks in the CI.
+  CMAKE_CONFIGURE_OPT: --disable-file-checks
+
 jobs:
 
   linux-multi:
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -13,6 +13,12 @@ on:
       - ".github/workflows/composite-cmake/**"
   pull_request:
 
+env:
+  # Due to busy file system problems on the CI runners, we introduced a patch to
+  # sleep if the MPI count of an I/O operation is unexpectedly 0 and then retry
+  # the I/O operation once. The time is in milliseconds.
+  IO_SLEEP_TIME: 5
```
cburstedde, line 20: Please prefix with SC_ for a consistent name space.

.github/workflows/ci.yml
```diff
@@ -37,23 +46,26 @@ jobs:
...
         make -j V=0
         make -j check V=0
 
     - name: Make check with MPI, without debug
       shell: bash
       run: |
         DIR="checkMPI" && mkdir -p "$DIR" && cd "$DIR"
-        ../configure --enable-mpi CFLAGS="-O2"
+        ../configure --enable-mpi CFLAGS="-O2" \
+        CPPFLAGS="-DSC_HACK_MPI_SLEEP=$IO_SLEEP_TIME"
```
cburstedde, line 59: Not to be too picky, but can you please indent this line by 4?

.github/workflows/ci_cmake.yml
```diff
@@ -112,6 +122,14 @@ jobs:
     - name: Checkout source code
       uses: actions/checkout@v4
 
+    - name: Patch I/O for CI by sleeping if I/O intensity is too high
+      run: git apply doc/patch/patch-CI-IO.patch
+
+    - name: Set macro for I/O waiting
+      shell: bash
+      # use an enviroment variable to prepend the flag to the bulit-in compiler flags
+      run: echo "CFLAGS=-DSC_HACK_MPI_SLEEP=$IO_SLEEP_TIME" >> "$GITHUB_ENV"
```
cburstedde, line 131: Should this be CPPFLAGS instead, or does CMake not use that variable?

doc/patch/patch-CI-IO.patch
```diff
@@ -0,0 +1,150 @@
...
+ 
++/** In case of an unexpected retrieved count 0, this macro sleeps and retries.
++ * In the Github Actions CI, we experienced transient CI failures due to the
++ * file system not performing requested I/O operations.
++ *
++ * The macro can be only callled in the case that expected count is unequal 0.
++ * It checks if count is 0 and if this is true, it sleeps and then retries the
++ * I/O operation once.
++ */
++#define SC_IO_SLEEP_AND_RETRY(func, time) do {\
```
cburstedde, line 40: I'd be leaning to prefer a function over a macro, quite generally. Would this be practicable here?

doc/patch/patch-CI-IO.patch
```diff
@@ -0,0 +1,150 @@
...
++                                      }\
++                                      retval = sc_io_error_class (mpiret,\
++                                                                  &errcode);\
++                                      SC_CHECK_MPI (retval);\
++                                      return errcode;\
++                                    }} while (0)
++
++
++/* This function is only needed if MPI IO is available */
++#ifdef SC_ENABLE_MPIIO
```
cburstedde, line 60: Is this kind of portability code replicated elsewhere in the code base? We may instead provide a helper function for this kind of sleep that is always available to all of libsc and libraries using it.

.github/workflows/ci.yml
tim-griesbach: Done

.github/workflows/ci.yml
tim-griesbach: Done

doc/patch/patch-CI-IO.patch
tim-griesbach: I moved the portable sleep function as `sc_sleep` to `sc.{c,h}`.

doc/patch/patch-CI-IO.patch
tim-griesbach: It is possible but would come with significantly more parameters. Namely, we would need `mpifile`, `offset`, `ptr`, `count`, `t` and `errcode` as additional parameters. What do you think about the trade off between using a macro but reducing the number of parameters from 8 to 2?

doc/patch/patch-CI-IO.patch
cburstedde: Cool!

.github/workflows/ci_cmake.yml
cburstedde: Ok. A bit sad, too...

doc/patch/patch-CI-IO.patch
cburstedde: Agreed, the use of the macro is rather elegant. I'm not sure if we should return out of it, would it make sense to return on the outside, or would that rather complicate things?

cburstedde:
It's a relief that this approach seems to resolve the issue with the CI file system. Maybe go for 10ms for safety? There was one remaining failure in the CI of the PR.

doc/patch/patch-CI-IO.patch
tim-griesbach: Yes, I adjusted the macro to return outside of it.

tim-griesbach:
> It's a relief that this approach seems to resolve the issue with the CI file system. Maybe go for 10ms for safety? There was one remaining failure in the CI of the PR.

After some testing of different sleeping times, I would propose to use 50 milliseconds since in my tests 10 milliseconds sometimes did not suffice but in yesterday's tests 5 milliseconds were fine. Given this variability of the occurrence of the CI fails, 50 milliseconds is maybe the more future proof option, which still results in reasonable CI runtimes.

tim-griesbach:
I think it would make sense to wait with merging this PR until tomorrow because the CI's behavior was so different today compared to yesterday. Hence, I would like to test again tomorrow to get an better idea of a suitable sleeping time.

tim-griesbach:
> What about merging now?

Sounds good to me. After more tests, I increased the I/O sleeping time again (now 0.5 seconds). Since my tests showed the behavior of the CI's file system is very much non-deterministic, I think the increase provides a more reliable CI and the impact on the overall CI time is still small.  

> Have you tested how this sc update works with the p4est CI, and if there's more to fix there (such as the patching etc.)?

This PR only changes functions that are not used in the p4est CI since the only parallel IO functions in p4est's CI are the legacy functions `sc_mpi_{read,write}`. I would propose to create a separate PR for this since firsts tests in p4est indicated that the higher I/O intensity in p4est tests may be an additional challenge.

cburstedde:
Ok, hoping the overall run time of the CI does not increase out of proportion due to the sleep calls.

Sounds good going after the p4est related functions next.

cburstedde:
As discussed earlier this week, we might forgo any attempt to purpose-code to the timings of the github runners' file systems. Maybe a simple configure option to deactivate the code of our tests that triggers this CI situation will work.

cburstedde:
This looks really good, thank you!

cburstedde:
Just puzzled, what's wrong with Windows this time?

tim-griesbach:
It seemed to be a SSL verification error during the download of the zlib code but I can not reproduce the error. Maybe it was due to a network problem of the Windows CI runner.

cburstedde:
Thanks! Let's proceed analogously in p4est.

PR #217, opened 2025-03-09
nary-payload: set payload in notify_nary for mpisize==1

hannesbrandt:
# Set payload in notify_nary for mpisize == 1 

This PR copies the content of in_payload to out_payload in notify_nary for the special case of mpisize == 1 and out_payload != NULL.

### DIFF
--- a/src/sc_notify.c
+++ b/src/sc_notify.c
@@ -1493,6 +1493,12 @@ sc_notify_payload_nary (sc_array_t * receivers, sc_array_t * senders,
       if (senders != NULL) {
         *(int *) sc_array_push (senders) = 0;
       }
+      if (out_payload != NULL) {
+        /* if out_payload is defined, copy single entry from in_payload */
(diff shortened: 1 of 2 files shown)


src/sc_notify.c
```diff
@@ -1493,6 +1493,12 @@ sc_notify_payload_nary (sc_array_t * receivers, sc_array_t * senders,
       if (senders != NULL) {
         *(int *) sc_array_push (senders) = 0;
       }
+      if (out_payload != NULL) {
+        /* if out_payload is defined, copy single entry from in_payload */
+        SC_ASSERT (in_payload != NULL && in_payload->elem_count == 1);
+        *(int *) sc_array_push (out_payload) =
```
cburstedde, line 1499: Would we be certain that on input the array length of out_payload is 0?
Also, the assertion on in_payload != NULL if out_payload != NULL might be stricter then previously.

cburstedde:
Thanks so much!

PR #218, opened 2025-03-09
Feature distro

cburstedde:
# Meet requirements for distribution

For some Linux distributions, sc_config.h must not include arch dependencies.

## Proposed changes

Since we really did not need most of these compiler and flag variables, remove them.
Any comments? Especially on the pkgconfig setup. @tamiko

### DIFF
--- a/src/sc.c
+++ b/src/sc.c
@@ -1293,9 +1293,6 @@ sc_init (sc_MPI_Comm mpicomm,
          int catch_signals, int print_backtrace,
          sc_log_handler_t log_handler, int log_threshold)
 {
-#if 0
-  const int           w = 24;
-#endif
   const char         *trace_file_name;
   const char         *trace_file_prio;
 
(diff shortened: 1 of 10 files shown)


cburstedde:
If @tim-griesbach would be able to double check that we don't break anything by removing these definitions?

src/sc.c
```diff
@@ -1293,7 +1293,9 @@ sc_init (sc_MPI_Comm mpicomm,
          int catch_signals, int print_backtrace,
          sc_log_handler_t log_handler, int log_threshold)
 {
-  int                 w;
+#if 0
+  const int           w = 24;
+#endif
```
tamiko, line 1298: The `#if 0` removes the `const int` definition. Wouldn't that lead to a compiler error in lines 1366ff where you use `w` in a macro?

tim-griesbach:
> If @tim-griesbach would be able to double check that we don't break anything by removing these definitions?

In my tests the code changes do not break anything. I pushed the small change of using the `volatile` keyword for endianness test, which we talked about.

src/sc.c
cburstedde: True, if we reset one of the two #if 0 to 1 and not the other, this will not work. This is a bit of lazy shortcut expecting that the two definitions apply to the same context. Maybe I should name the define. Let me do that.

cburstedde: Ok I just killed it.

PR #219, opened 2025-03-10
scda: Rename options to params and add logging level

tim-griesbach:
# scda: Rename options to params and add logging level

This PR renames the scda options structure to `sc_scda_params_t` following the same semantics as the params structures in `p4est_wrap.h` and `p4est_mesh.h`. This includes introducing a new function to initialize the parameters, namely `sc_scda_params_init`. Additionally, this PR adds the logging level to the scda parameter structure and the scda file context to enable the user to control the logging of I/O errors.


cburstedde:
Thanks!

PR #221, opened 2025-03-13
Feature visibility

cburstedde:
# Add visibility attribute for GCC >= 4

## Proposed changes

Using p4est from a certain python library turned up multiple instances of the same extern variable, namely the package id (used internally for registering log priorities and debug memory counters). Add some defines to use the visibility attribute.

The attribute is placed only in the .h file for now. To make it work on windows, we may have to add it in front of the definition in the .c file, which I find a bit annoying. Shall we do it anyway or is the current approach sufficient?

Curious on feedback. @tamiko @sloede @lukasdreyer

### DIFF
--- a/doc/release_notes.txt
+++ b/doc/release_notes.txt
@@ -26,6 +26,7 @@
  - Flag the scda API to be changed in the next release
  - Intremental documentation updates
  - Add a portable sleep function
+ - Remove unused sc_handler_t
 
 ### Build system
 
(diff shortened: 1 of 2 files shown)


cburstedde:
I'll merge this (possibly partial) fix now since we're getting ready for a release. I'll be more than happy to reopen for the cygwin and win32 branches in sc.h.

PR #226, opened 2025-07-17
Do not mark functions in header files as 'static'.

bangerth:
Functions marked as `static` are local to a translation unit, and consequently if you mark a function in a header file as `static`, it is replicated into each file that `#include`s the header file. If the function is marked as `inline`, this would ordinarily give the linker the ability to unify all of the copies of that function, but if the function is also `static`, you're not giving the linker the ability to do this. As a consequence, functions in header files *should* be marked as `inline`, but not as `static`.

This patch fixes this for the one file in libsc where that happens. This is necessary also to make it possible to build C++20 modules, see https://github.com/dealii/dealii/issues/18071.

### DIFF
--- a/src/sc_containers.h
+++ b/src/sc_containers.h
@@ -492,7 +492,7 @@ size_t              sc_array_pqueue_pop (sc_array_t * array,
  * \return           Pointer to the indexed array element.
  */
 /*@unused@*/
-static inline void *
+inline void *
 sc_array_index (sc_array_t * array, size_t iz)
 {
   SC_ASSERT (iz < array->elem_count);
@@ -507,7 +507,7 @@ sc_array_index (sc_array_t * array, size_t iz)
  *                   NULL if the specified index is elem_count.
  */
 /*@unused@*/
-static inline void *
+inline void *
 sc_array_index_null (sc_array_t * array, size_t iz)
 {
   SC_ASSERT (iz <= array->elem_count);
@@ -520,7 +520,7 @@ sc_array_index_null (sc_array_t * array, size_t iz)
  * \param [in] index needs to be in [0]..[elem_count-1].
  */
 /*@unused@*/
-static inline void *
+inline void *
 sc_array_index_int (sc_array_t * array, int i)
 {
   SC_ASSERT (i >= 0 && (size_t) i < array->elem_count);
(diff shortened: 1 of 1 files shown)


cburstedde:
Thanks, I'd share this observation. Back then this design was safest considering compilers respecting inline or not, but it may be badly outdated now.

Just wondering about the CI issues -- it may be that some files like sc_options.c need an explicit `#include <sc_containers.h>`.

And please if I may ask to add the usual doc/author_<lastname>.txt to the sc library as well.

bangerth:
I added the author license release. As for the CI errors, I am confused. They are all of the form
```
/usr/bin/ld: src/.libs/libsc.a(libsc_la-sc_options.o): in function `sc_options_destroy_internal':
/home/runner/work/libsc/libsc/checkMPIdebug_valgrind/../src/sc_options.c:287:(.text+0x7af): undefined reference to `sc_array_index'
/usr/bin/ld: /home/runner/work/libsc/libsc/checkMPIdebug_valgrind/../src/sc_options.c:309:(.text+0x8ce): undefined reference to `sc_array_index'
/usr/bin/ld: src/.libs/libsc.a(libsc_la-sc_options.o): in function `sc_options_add_item':
/home/runner/work/libsc/libsc/checkMPIdebug_valgrind/../src/sc_options.c:348:(.text+0xa8d): undefined reference to `sc_array_push'
/usr/bin/ld: src/.libs/libsc.a(libsc_la-sc_options.o): in function `sc_options_add_suboptions':
/home/runner/work/libsc/libsc/checkMPIdebug_valgrind/../src/sc_options.c:530:(.text+0x10d1): undefined reference to `sc_array_index'
```
In other words, these files have all *compiled* fine, which means that the compiler has successfully seen a declaration of the functions in question, but apparently has not created a definition. I take it that that means that each file that was compiled has (directly or transitively) included `sc_containers.h` somehow -- otherwise, the compiler would have had to find a declaration somewhere else, but the one in `sc_containers.h` is the only one I can find in the libsc source base. 

But the linker then complains. That would mean that the compiler sees a *declaration but not a definition*. I don't know how that can be, since the two are provided in one. Do you have ideas?

cburstedde:
In fact, it would be sufficient to cherry-pick 9a26c29cce5ea8a35d23ea6c3901cfb042352c57.

bangerth:
I believe so.  I think this was the only file we tripped over.

PR #230, opened 2025-09-15
Remove hard-coded CMAKE_C_STANDARD

dutkalex:
# Remove hard-coded `CMAKE_C_STANDARD`

Proposed changes:
CMake globals (i.e. `CMAKE_XXX` variables) should not be modified by the CMakeLists.txt, as this breaks user workflows and causes problems for downstream projects. This PR fixes this in the case of the `CMAKE_C_STANDARD` variable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,21 +15,23 @@ include(GNUInstallDirs)
 include(cmake/config.cmake) # options.cmake before config.cmake to determine user intent
 include(cmake/compilers.cmake) # compilers.cmake must be AFTER find_package() calls in config.cmake in general
 
-set(CMAKE_C_STANDARD 99)
 message(STATUS "libsc ${PROJECT_VERSION} "
                "install prefix: ${CMAKE_INSTALL_PREFIX}")
 
 # --- iniparser
 add_library(iniparser OBJECT iniparser/inistring.c iniparser/dictionary.c iniparser/iniparser.c)
+target_compile_features(iniparser PUBLIC c_std_99)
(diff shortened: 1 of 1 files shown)


dutkalex:
> I would suggest rather:
> 
> ```cmake
> if(NOT DEFINED CMAKE_C_STANDARD)
>     set(CMAKE_C_STANDARD 99)
> endif()
> ```
>
> if a downstream project already set the variable, use it, else use C99 as a default value.

I guess this approach would be somewhat equivalent since noone in their right mind would ask for anything less than C99 today (you wouldn't get any diagnostic however and it would crash at compile time rather than at configure time though).
I would ague that the modern target-based approach is better because it gets rid of the global state in favor of declarative target-specific properties which can be reasoned about locally both by someone trying to debug, as well as by cmake (to produce clearer error messages for example).
Generally speaking, the CMakeLists is not the place to enforce such kinds of global configuration parameters, because of the lack of context. The right answer is probably different for an isolated build, for one within a super-project, for one triggered by a PM, for one within a CI pipeline, etc. and in the end, the calling code (or the final user) is the only one who can provide that context. CMake's job is therefore to *check* that the configuration satisfies the expressed requirements. If you try to enforce such globals, you are breaking someone's code or workflow downstream, because there is no "one size fits all" answer. I can't highly enough recommend [that talk by Brett Brown](https://www.youtube.com/watch?v=xNHKTdnn4fY) on this matter. 

> If you really prefer your way of doing it, perhaps creating a cmake macro to avoid duplicating `target_compile_features` at several locations would be better.

Besides this being a feature rather than a bug IMO, I don't quite get the point of creating a macro here which would also amount to 1 line per target. Maybe I didn't understand your comment correctly...

PR #236, opened 2026-02-08
Fix MSVC-specific syntax issues

ZHBHFUT:
# Fix MSVC-specific syntax issues

Following up on issue #232 .

This PR refines the previously proposed platform-independent annotation macros and updates the affected declarations to better cooperate with the existing `scindent` workflow.

Changes since the previous revision:
- Adjusted the placement and formatting of the new macros to avoid confusing GNU indent.
- Verified that `scindent` runs cleanly on the touched headers without warnings or errors.
- No functional changes compared to the earlier version; this update is strictly about formatting robustness and tool compatibility.

The goal remains to provide a unified, portable way to express attributes such as `noreturn` and `printf`-like annotations across compilers, while keeping the resulting code friendly to the current formatting infrastructure.

Please let me know if you would prefer a different placement or style for any of the annotations.

### DIFF
--- a/example/pthread/pthread.c
+++ b/example/pthread/pthread.c
@@ -21,6 +21,12 @@
   02110-1301, USA.
 */
 
+#ifdef _MSC_VER
+#define WIN32_LEAN_AND_MEAN
+#include <Windows.h>
+#define sleep Sleep
+#endif
+
 #include <pthread.h>
 #include <sc_options.h>
 
--- a/iniparser/inistring.h
+++ b/iniparser/inistring.h
@@ -36,6 +36,19 @@
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE             /**< Enable additional functionality. */
 #endif
+
+/* ​Platform-specific printf format string checks​​ */
+#ifndef SC_PRINTF_LIKE
+#ifdef _MSC_VER
+#include <sal.h> /* _Printf_format_string_ */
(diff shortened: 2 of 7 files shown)


ZHBHFUT:
For Windows/MSVC users, here is a minimal CMake + vcpkg configure/build/install sequence (x64):
```cmd
@echo off
setlocal

echo --- Configure (Visual Studio generator) ---
set "VCPKG_ROOT=D:\OpenSource\vcpkg"

cmake -S . -B build ^
  -A x64 ^
  -DCMAKE_INSTALL_PREFIX="%CD%\build\install" ^
  -DSC_ENABLE_MPI=ON ^
  -DCMAKE_TOOLCHAIN_FILE="%VCPKG_ROOT%\scripts\buildsystems\vcpkg.cmake" ^
  -DCMAKE_C_FLAGS="/utf-8 /wd4100 /wd4702 /wd4267"

echo --- Build ---
cmake --build build --config Release

echo --- Install ---
cmake --install build --config Release

endlocal
```
Additional options for building shared libs (DLL) if supported by the project/config:
```cmd
  -DBUILD_SHARED_LIBS=ON ^
  -DSC_BUILD_SHARED_LIBS=ON ^
  -DCMAKE_C_FLAGS="/DBUILDING_DLL /utf-8 /wd4100 /wd4702 /wd4267"
```

Note that depending on the checkout location and local filesystem permissions, the install step may require elevated privileges on Windows, even when installing into a subdirectory of the build tree.

